PR #12, opened 2026-06-08
perf: reduce collector CPU usage with smarter defaults and throttling

SaraiNoQ:
The collector was causing excessive CPU load through several compounding issues, especially noticeable when actively using AI coding tools:

1. allTimeSince defaulted to 2024-01-01, causing every --since call to scan 2+ years of history. Default now auto-discovers the earliest local activity by scanning AI tool data directory creation times (one-time, persisted to settings.json). Falls back to 90 days.

2. Chokidar file watcher polled 9 directories every 2 seconds. Increased to 10s — watcher is a bonus mechanism for faster-than- interval updates, not the primary collection path.

3. File-change debounce was 1.5s, triggering heavy tokscale spawns nearly continuously during active coding. Increased to 10s.

4. Watcher-triggered ticks now skip the expensive --since allTimeSince call entirely — only today/month are collected on file changes. The full allTime scan runs exclusively on the 5-minute interval tick, avoiding redundant full-history scans.

5. Increased awaitWriteFinish stability threshold from 500ms to 2000ms to avoid collecting mid-write files.

Overall expected CPU reduction: ~80-90%, particularly during active AI tool usage where file-watcher churn was the dominant cost.

### DIFF
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -5,7 +5,7 @@ const path = require('node:path');
 const { defaultDeviceId, loadDotEnv, parseArgs, pidFilePath } = require('../shared/config');
 const { appVersion } = require('../shared/appVersion');
 const { clientsCsvForSetting } = require('../shared/clientTracking');
-const { collectUsageOnce, startCollector } = require('../shared/collector');
+const { collectUsageOnce, startCollector, discoverEarliestActivity } = require('../shared/collector');
 const { normalizeLimitsRefreshMs, parseBoolean, parseLimitProviders } = require('../shared/limitCollector');
 
 loadDotEnv();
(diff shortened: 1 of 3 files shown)


Javis603:
Thanks for this — the profiling instinct is appreciated, and a couple of these point in a useful direction. I dug into tokscale's internals before deciding, and that reframed how I read the PR, so here's where I land on each piece.

**Concurrency** — already addressed on `main`: the today/month/all-time calls now run in parallel instead of sequentially ([d64b027](<https://github.com/Javis603/token-monitor/commit/d64b027>)), so wall-clock is `max()` of the three rather than `sum()`.

**Skipping all-time on watcher ticks (points 1 & 4)** — these both rest on the idea that an earlier `--since` scans more history, so I checked tokscale's source. That premise doesn't hold:
- The date filter (`--today` / `--since`) is applied in the aggregator *after* every file is loaded — tokscale reads all session files regardless of the date range and filters by timestamp at the end. So `--since 2024-01-01` and `--today` do the same scan work; there's no "2+ years of history" that a later date avoids.
- It keeps a persistent per-file message cache (`source-message-cache.bin`), so unchanged files are never re-parsed. The remaining per-run cost is the cache-validation fingerprint — a full read + SHA-256 per file, with no mtime fast-path — and that's paid equally by all three calls.

So all-time isn't the expensive call; all three cost roughly the same. Skipping it on watcher ticks does cut ~⅓ of redundant I/O, but since the three calls are now parallelized it buys ~0 wall-clock, and idle CPU here measures ~2%. I don't think the `skipAllTime` branch earns its keep right now — though "all three calls re-scan the same files" is the real cost worth keeping in mind if this is ever revisited.

**`allTimeSince` auto-discovery (point 1)** — I'd prefer to leave the default as-is. It changes the semantics of the "All Time" total (and per-client detection status derives from it), the 90-day fallback can silently shrink existing users' numbers, and `birthtime` is unreliable on Linux and several filesystems. Combined with the point above — an earlier `--since` doesn't actually add scan cost — the CPU upside is small relative to the risk.

**Polling, debounce, and awaitWriteFinish (points 2, 3, 5)** — polling itself is intentional here (native FS events were tried and reverted for reliability), so the knob is the interval, not the mechanism. 10s polling + 10s debounce stacks into ~20s worst-case latency before a change shows up, which is a noticeable hit for a "live" widget.

Net: I'm not merging as-is, but the writeup was genuinely useful and pushed me to verify how the collector actually behaves. Really appreciate you taking the time to dig in 🙏

Javis603:
Following up now that [v0.11.1](<https://github.com/Javis603/token-monitor/releases/tag/v0.11.1>) is out, because it changes two things I said above.

First, a correction: the parallel scans I pointed to turned out to amplify a worse problem. Concurrent spawns tripled the peak load and, on macOS, could exhaust syspolicyd's fds and break Gatekeeper checks for unrelated apps (#15), so the three scans are serial again: [313cd3f](<https://github.com/Javis603/token-monitor/commit/313cd3f78ce6d7c6420dc58fe90a1aa18cdb9fc7>). The bigger CPU story behind that issue was a self-trigger loop (the watcher monitored cache dirs our own sync writes), details in #15.

Second, the direction behind your points 1 and 4, making watcher ticks cheaper than a full rescan, did land in a different form: watch-triggered ticks now run a single `--today` scan, and the month / all-time figures are derived exactly from the last full scan as an anchor delta, so nothing goes stale or estimated. Full scans only run on startup, the 5-minute interval, and manual refresh: [f9e1cdb](<https://github.com/Javis603/token-monitor/commit/f9e1cdbd6b4db3315afa4533006202e18017ffa4>). Your observation that all three calls re-scan the same files is exactly the cost this removes.

The remaining items here (allTimeSince auto-discovery, 10s polling/debounce, awaitWriteFinish tuning) are the parts I'd already passed on, so I'll close this one out. Thanks again for the profiling, it held up well 🙏

PR #21, opened 2026-06-20
feat: edge dock — drag window to screen edge to auto-hide

xxtt-01:
## Summary

Adds edge dock support — drag a window to the screen edge and it auto-hides with only a 5px strip visible. Hover the strip to slide it back. Windows only.

## Changes

**Configuration** (`main.js`, `index.html`):
- New `settings.edgeDock` toggle in Display settings
- Auto-restores dock position across app restarts

**Edge dock state machine** (`main.js`):
- `edgeDo` / `edgeUndo` — enter/exit dock mode with position tracking
- `edgeAfterMoved` — state-guard pattern (tracks dock/expand positions, detects user drag-away)
- `edgeStartMonitor` — adaptive cursor polling (150ms near edge, 500ms backoff when idle)
- `edgeSlide` — easeInOutQuad animation, 8 steps ~60ms
- `win.on('move')` — disables resize near edge to prevent Windows Snap interference

**UI** (`app.js`, `styles.css`):
- Settings checkbox with `data-i18n` translation support
- Edge glow indicator when docked (CSS animation, injected via IPC)
- Platform-gated: only visible and active on Windows

**IPC** (`preload.js`):
- `onEdgeDockState` bridge for real-time dock state in the renderer

## Review items addressed

All feedback from @Javis603 has been addressed:

1. **i18n trailing comma** (blocker) — fixed the missing comma on zh-TW `windowShortcutConflict` line
2. **Platform restriction** — edge dock is now Windows-only: settings checkbox hidden on macOS, mount logic and event handlers guarded with `process.platform === 'win32'`
3. **CPU backoff polling** — changed from fixed 150ms `setInterval` to adaptive recursive `setTimeout`: 150ms when cursor is near the dock edge, 500ms when idle

Additional bugs found and fixed during self-review:
- Fixed `atDock` ReferenceError (misplaced const inside try block)
- Fixed non-Win32 hiding (used classList.add('hidden') without a matching CSS rule)
- Fixed glow element positioning (left/right was inverted)

## Files changed

| File | Change |
|------|--------|
| `src/electron/main.js` | Edge dock state machine + platform guards + CPU backoff |
| `src/electron/preload.js` | `onEdgeDockState` IPC bridge |
| `src/electron/renderer/app.js` | Settings UI + glow indicator + platform gate |
| `src/electron/renderer/index.html` | Settings checkbox |
| `src/electron/renderer/i18n.js` | Translation keys (en/zh-TW/zh-CN) |
| `src/electron/renderer/styles.css` | Glow animation keyframes |

Thanks for the careful review — the CPU profiling and platform testing were especially helpful.


Javis603:
Thanks for splitting these out, three separate features are much easier to follow.

I tried this on both platforms and ran into two different problems:

On macOS it doesn't behave. The window keeps getting pulled back: macOS's WindowServer constrains windows from moving off-screen, so a `setBounds` to the dock position (only 5px showing) gets corrected back, and it takes many drags before it occasionally sticks. This doesn't look like something a parameter tweak fixes; doing it properly on mac probably means going to the native layer (something like NSPanel / collectionBehavior, a non-activating panel that's allowed off-screen), which is a different level of effort from the current pure-`setBounds` approach.

On Windows it works, but idle CPU jumps noticeably. On a 9950X3D the idle baseline is 0.2%, and it spikes to 3% during the dock interaction. The cause is the 150ms cursor polling that runs while docked (plus the slide animation ticking roughly every 7.5ms). This project is sensitive to idle CPU (issue #15 was a busy-loop burning CPU), so before merging I'd like this tightened up: while docked, switch to backoff polling (widen the interval when idle), or at least loosen that 150ms substantially.

On balance I'm inclined to limit this feature to Windows. Edge-docking is a very Windows interaction, macOS has no equivalent system convention for it, and it genuinely doesn't run well on mac. Concretely, wrap both the mount logic and the settings checkbox in `process.platform === 'win32'` so it doesn't show up on mac at all, and tighten the Windows CPU above.

One hard blocker right now: in `src/electron/renderer/i18n.js`, the `windowShortcutConflict` line (743) is missing a trailing comma, so the next line `edgeDock` throws `SyntaxError: Unexpected string`, `node --check` fails, and once i18n breaks the whole renderer won't load (`i18n.test.js` fails entirely because of this too). Fix that first.

xxtt-01:
Thank you for the thorough review and for testing on both platforms. All three issues have been addressed in the latest commit (0b1f501):

**1. i18n trailing comma** (blocker) ✅
Fixed — the zh-TW `windowShortcutConflict` line now has its trailing comma. `node --check` and all `i18n.test.js` tests pass.

**2. Platform restriction** ✅
Edge dock is now Windows-only:
- The settings checkbox is hidden on non-Win32 (`style.display = 'none'`)
- `startEdgeDock()` in `applyWindowSettings` is guarded with `process.platform === 'win32'`
- The `moved` handler's `edgeAfterMoved()` call and the `move` handler's Snap prevention logic are both gated

**3. CPU backoff polling** ✅
Replaced the fixed 150ms `setInterval` with an adaptive recursive `setTimeout`:
- 150ms (EDGE_POLL_MS) when the cursor is near the dock edge or the window is expanded
- 500ms (EDGE_POLL_BACKOFF_MS) when docked and the cursor is away from the edge
- Animation timing (~7.5ms per tick, 60ms total for 8 steps) is unchanged

During self-review I also found and fixed three additional bugs:
- A `ReferenceError` where `atDock` was declared with `const` inside the try block but referenced outside it for the adaptive delay calculation
- Non-Win32 hiding used `classList.add('hidden')` but this branch doesn't have a generic `.hidden` CSS rule — fixed by using `style.display = 'none'` instead
- The glow element's left/right positioning was inverted, showing the dock indicator on the wrong edge

Rebased onto the latest upstream/main. Thanks again for the detailed feedback.

xxtt-01:
自审发现并修复了一个额外问题：顶部贴边的光效位置错误（ → ）。贴边到屏幕顶部时，光效条会出现在屏幕底部，现已修正。提交 5376f59。

PR #23, opened 2026-06-20
feat: startup optimization — anchor persistence, dir timestamp cache, progressive loading

xxtt-01:
## Summary

This PR optimizes startup time by persisting anchor data between restarts and skipping redundant tokscale scans when no data directories have changed.

### Changes

**Anchor persistence** (`collector-anchor.json`):
- After a full scan, the aggregated month/allTime periods are written to disk
- On subsequent launches, the anchor is loaded and validated by `dateKey`
- When valid, only `--today` is scanned; month/allTime are derived via `applyPeriodDelta`
- Empty anchors (zero data from a failed scan) are detected and trigger a fresh scan

**Directory timestamp caching** (`collector-dirts.json`):
- After a full scan, mtime of every file inside each client data directory is collected
- On launch, current mtimes are compared; unchanged → entire tokscale scan is skipped
- Uses **file-level mtime** (not directory mtime) for cross-platform reliability — on APFS and NTFS, appending to a file does not update the parent directory's mtime
- Self-synced clients (Cursor, Antigravity) are excluded from the check so their periodic syncs don't invalidate the cache

**Progressive data push**:
- `onProgress` callback sends partial results (today → month → allTime) to the renderer as each period completes
- The UI shows data incrementally instead of waiting for all three scans

**Sync reliability fix**:
- `maybeSyncCursor` / `maybeSyncAntigravity` moved before the `skipTokenscan` guard so syncs always run (rate-limited internally to 5min)

- macOS/APFS directory mtime issue: Changed to file-level mtime collection, which works correctly on all platforms

### Commits

- `b80a63c` — feat: startup optimization (77 insertions, 9 deletions) + file-level mtime fix

### DIFF
--- a/src/shared/collector.js
+++ b/src/shared/collector.js
@@ -363,16 +363,22 @@ async function collectUsageOnce(options) {
   let month = emptyPeriod();
   let allTime = emptyPeriod();
   const anchor = options.todayOnlyAnchor;
-  const anchorUsed = Boolean(anchor && anchor.dateKey === localTodayKey());
   if (normalizedClients) {
     await maybeSyncCursor(normalizedClients, options.logger);
     await maybeSyncAntigravity(normalizedClients, options.logger, options.homeDir || os.homedir());
-    if (anchorUsed) {
+
+    if (options.skipTokenscan) {
+      today = anchor.today;
+      month = anchor.month;
+      allTime = anchor.allTime;
+      options.onProgress?.({ today, month, allTime, updatedAt: new Date().toISOString() });
+    } else if (anchor && anchor.dateKey === localTodayKey()) {
       // Anchored tick (watch-triggered): every tokscale period scan costs the
       // same full load + filter, so scan only --today and update the broader
       // windows exactly via applyPeriodDelta — one spawn instead of three.
       const todayJson = await runTokscaleFn({ clients: normalizedClients, flags: ['--today'], commandTimeoutMs });
       today = extractUsageFromTokscale(todayJson);
+      options.onProgress?.({ today, month: null, allTime: null, updatedAt: new Date().toISOString() });
       month = applyPeriodDelta(anchor.month, today, anchor.today);
       allTime = applyPeriodDelta(anchor.allTime, today, anchor.today);
     } else {
@@ -382,7 +388,9 @@ async function collectUsageOnce(options) {
       const monthJson = await runTokscaleFn({ clients: normalizedClients, flags: ['--month'], commandTimeoutMs });
       const allTimeJson = await runTokscaleFn({ clients: normalizedClients, flags: ['--since', allTimeSince], commandTimeoutMs });
       today = extractUsageFromTokscale(todayJson);
+      options.onProgress?.({ today, month: null, allTime: null, updatedAt: new Date().toISOString() });
       month = extractUsageFromTokscale(monthJson);
+      options.onProgress?.({ today, month, allTime: null, updatedAt: new Date().toISOString() });
       allTime = extractUsageFromTokscale(allTimeJson);
     }
     applySessionTimestamps({ today, month, allTime }, options.homeDir || os.homedir());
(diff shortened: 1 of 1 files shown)


Javis603:
The direction is right. Persisting the anchor, reusing the last full scan across restarts, and skipping that startup full scan: I'm on board with the idea. Switching the directory check from directory mtime to file-level mtime is also correct; you nailed the fact that on APFS/NTFS appending to a file doesn't update the parent directory's mtime.

Going through the source and tests, though, a few things aren't mergeable yet:

1. The skip only scans one level deep, so it misses real changes. `collectDirTimestamps` records each data directory's own mtime plus its immediate children's, but these clients store data nested: `~/.claude/projects/<project>/x.jsonl`, `~/.codex/sessions/<year>/<month>/<day>/rollout.jsonl`. On APFS, appending to an existing nested file updates neither the parent directory's mtime nor anything in the level you record, so `timestampsEqual` stays true and `skipTokenscan` serves stale data mid-session or after a restart in the middle of one, which is exactly the 3 to 5 second live-update promise. Either recurse the scan down to files (take the max mtime), or just keep the persisted anchor and drop the skip entirely.

2. WSL totals come out wrong (Windows+WSL). The anchor used to store `captured.windowsPeriods`, i.e. Windows-only, before the WSL merge. You changed it to store `summary.today/month/allTime` (post-merge) and also removed `wslAnchor = captured.wslBundle`, so `wslAnchor` is always null. As a result the anchored / today-only tick no longer re-merges WSL: `today` loses the WSL portion, and `month = applyPeriodDelta(mergedMonth, windowsToday, mergedToday)` subtracts the WSL amount out of month. The today-only delta is an identity that requires the anchor and the scan result to be in the same space (both Windows-only); a merged anchor breaks that premise. This needs the anchor kept Windows-only, with wslAnchor stored separately and re-merged on the anchored tick.

3. Progressive updates break the carry-forward contract and tend to flash empty. onProgress explicitly passes `history: null` and `limits: null`, but `carryDeviceHistory`'s contract is to carry forward only when the incoming object omits the history key (`if (hasOwn(incoming,'history')) return incoming`). Setting the key to null bypasses it, so the renderer gets an explicit null on every full tick and trends/limits tend to blink empty then come back. Just omit those two keys on the progressive pushes.

4. The persisted anchor isn't tied to config. It stores `{dateKey,today,month,allTime}` without clients / allTimeSince, so after you change the tracked clients or the start date, the next startup keeps computing off the old baseline. Either fold those into the key, or invalidate and rescan when the config changes.

5. Tests. Running locally here, the debounce test in `collectorLoadGuards` fails, `collectorTodayDelta` never finishes and hangs (killed after 20s), and the new persisted anchor / dir skip / progressive push have no tests at all.

This one touches the collector core plus your WSL delta, so the surface is large and the bar is highest. The progressive-loading part is relatively independent and lowest-risk; if you want a smaller one to land first, you could split that out (after the null-key fix). The anchor persistence + dir skip set involves recursive mtime / WSL / config scope, so I'd suggest reworking that as a fresh pass.

xxtt-01:
感谢细致的审查！根据你的建议，我决定将这个 PR 拆分为两个独立的部分：

**1. 渐进推送**（已提交为新 PR）— 修复了 `onProgress` 的 null key 问题，同时将渐进推送拆为独立的、风险较低的功能。

**2. Anchor 持久化 + dir-skip** — 需要重新处理递归 mtime、WSL anchor 隔离和配置绑定，将单独重新设计。

这个 PR 先关闭，拆分后的工作会通过新的 PR 提交。再次感谢你的时间！

xxtt-01:
As mentioned in the close comment, the progressive-loading portion has been split out into PR #26 (https://github.com/Javis603/token-monitor/pull/26). The null-key issue from review item #3 is fixed there — history and limits are now omitted from the partial payload instead of being set to null, so carryDeviceHistory works correctly and the UI no longer flashes empty.

The remaining items (dir depth, WSL anchor, config binding, tests) will be addressed in a follow-up PR after rework.

PR #24, opened 2026-06-20
feat: OpenCode multi-account — per-profile rate-limit tracking

xxtt-01:
## Summary

Extends OpenCode account support from a single cookie to multiple named profiles, each with independent rate-limit tracking.

### Configuration
- New `settings.opencodeProfiles` object stores named profiles: `{ "work": { cookie: "auth=...", enabled: true } }`
- Legacy `settings.opencodeCookie` auto-migrated to `profiles.default` on read
- Environment variable `TOKEN_MONITOR_OPENCODE_COOKIE` continues to work as a special `(env)` entry

### Collection (`limitCollector.js`)
- `fetchOpenCodeLimits` uses dual-path: 1 cookie → legacy merged behaviour, 2+ cookies → per-profile parallel queries
- `fetchSingleOpenCodeProfile` with 15s `Promise.race` timeout
- `providerCollapseKey` (`limits.js`) prevents profile aggregation via `providerAggregateKey`

### IPC (`main.js`)
- 5 new handlers: `getProfiles`, `saveProfile`, `deleteProfile`, `renameProfile`, `setProfileEnabled`
- `opencode:saveCookie` / `opencode:logout` / `opencode:status` updated for profiles
- `opencode:getProfiles` strips cookie values for security (renderer only needs name/enabled)
- `opencode:status` queries all profiles in parallel

### UI (`app.js`, `index.html`, `styles.css`)
- Profile list with enable/disable toggle, inline rename, delete
- Add profile form with step-by-step instructions
- Per-profile rate-limit bars in the limits view
- Full i18n support across en / zh-TW / zh-CN

### Account key
- Derived from the account itself (`go:{workspaceId}` / `zen:{workspaceId}`) or a stable SHA-256 hash of the cookie as fallback — never from the user-editable profile name, so the same account is consistently identified across machines and renames

## Review items addressed

All feedback from @Javis603's review has been addressed:

1. **accountKey stability** — changed from `profile:{name}` to workspaceId/cookie-hash based keys, matching the single-account path
2. **Tests** — `cursorSettingsLayout.test.js` updated for the new panel structure; 366 tests passing
3. **Local card removed** — multi-account mode no longer emits a standalone "Local" provider (it was always a single-account fallback; redundant once cookies are configured)
4. **Security** — `opencode:getProfiles` strips cookie values before returning to the renderer
5. **i18n** — all previously hardcoded Simplified Chinese strings replaced with `t()` calls; 17 new keys across en/zh-TW/zh-CN (51 entries)
6. **Parallel queries** — both `fetchOpenCodeLimits` and the `opencode:status` handler now use `Promise.all` instead of serial iteration

Additional self-review fixes:
- `renderOpenCodeAccountGroup` hardcoded `"accounts"` string → i18n key `settings.opencode.nAccounts`
- Removed dead `refreshOpencodeStatus` function
- OpenCode toggle now uses `setAccountGroupExpanded` shared pattern (consistent with Cursor/Codex/DeepSeek panels)

Rebased onto latest upstream/main (includes `dcde9da` session cookie redaction, `01a3237` home overview).

## Files changed

| File | Change |
|------|--------|
| `src/shared/limitCollector.js` | Dual-path + parallel per-profile fetching with stable accountKey |
| `src/shared/limits.js` | `providerCollapseKey` for opencode |
| `src/electron/main.js` | Profile settings, 5 IPC handlers, auto-migration, parallel status |
| `src/electron/preload.js` | Profile API bridge |
| `src/electron/renderer/app.js` | Profile list UI, limit display group, i18n cleanup |
| `src/electron/renderer/index.html` | Profile panel with i18n |
| `src/electron/renderer/styles.css` | Profile styles |
| `src/electron/renderer/i18n.js` | 17 keys × 3 locales |
| `tests/electron/cursorSettingsLayout.test.js` | Updated for new panel structure |

Thank you for the thorough review — the feedback on accountKey stability and the multi-account local/web model was particularly helpful in getting the design right.


Javis603:
This is the closest to mergeable of the three. The multi-account work reuses the codex setup directly (`accountKey`/`accountLabel`, with `providerCollapseKey` going through `providerAggregateKey`), and `limits.js` is a one-line change: the wiring is right. `fetchOpenCodeLimits` returning an object for single-account and an array for multi-account had me worried, but `collectLimitsOnce` flattens it with `Array.isArray`, so that's fine.

A few things to fix before merging:

1. The accountKey from the profile name is unstable and will cross accounts. `accountKey = hash('opencode', 'profile:'+name)`, and opencode then aggregates by that key across devices. Two machines both named `default` but logged into different OpenCode accounts collide on the key, so the hub overwrites one with the other; the same account named differently on two machines gets split into two rows. This is data correctness, not a display thing. Derive it from the account itself: your single-account path already uses `go:${goWeb.workspaceId}` / `zen:${...}`, so reuse the same source for multi-account, or failing that a stable hash derived from the cookie, rather than the user-entered label.

2. Three existing tests fail. This reworks the OpenCode settings panel and replaces the old single-account `opencodeLogoutButton` structure, but the tests pinning the old structure weren't updated, so `cursorSettingsLayout.test.js` has 3 failures (the OpenCode panel one plus two DeepSeek). Updating them to the new structure is all it takes.

3. Multi-account breaks the single-account local/web model. Single-account is web-preferred and falls back to local only when web is unavailable, chosen wholesale. Here multi-account turns local into a permanent standalone "Local" card while each profile runs web-only, and shows both at once. The issue is that the local data from `collectGo` is the one account logged in locally (its identity is `opencode-go:<db path>`, not "all accounts summed", and that comment should be fixed), and it's in a different namespace from web's `workspaceId`, so there's no way to tell whether it's the same account. So if your local CLI is logged into one of the web profiles' accounts, you'll see the "Local" card and that profile's card showing the same usage twice; and since `fetchSingleOpenCodeProfile` is web-only, a profile whose web call fails has no local fallback. I'd suggest dropping the standalone Local card in multi-account mode (local was only ever the single-account fallback for when there's no web cookie, so once you've added cookies it's redundant) and keeping per-profile web.

A few smaller things:

- On security, the new `opencode:getProfiles` returns each profile's raw cookie to the renderer. The UI only needs the name / enabled state / connected status, not the cookie itself, so I'd strip the cookie in the handler before returning.
- The new OpenCode UI has ~10 hardcoded Simplified Chinese strings (`尚未添加任何账号` / `已禁用` / `重命名` / `确定删除账号「…」吗?` / `已连接` etc.) that bypass i18n, so English / Traditional Chinese users see Simplified; there's also a native `confirm()` dialog that doesn't quite fit the existing style. Routing these through i18n would be more consistent.
- Profiles are fetched serially with a 15s timeout each, so startup gets slow with several accounts; worth considering fetching them in parallel.

Once these are sorted I think it's in good shape

xxtt-01:
Thank you for the detailed and thoughtful review @Javis603. All six items have been addressed in the latest commit (1469fc1). Here's a summary of each fix:

**1. accountKey stability** ✅
Changed from `profile:{name}` to derive from the account itself: `go:{workspaceId}` / `zen:{workspaceId}` when available, falling back to a stable SHA-256 hash of the cookie (`cookie:{hash12}`). This matches the single-account path and is consistent across renames and different machines.

**2. Tests** ✅
`cursorSettingsLayout.test.js` updated for the new profile panel structure. The DeepSeek test references that depended on `renderOpencodeStatus` now point to `renderOpenCodeProfiles`. All 366 tests pass.

**3. Local card in multi-account mode** ✅
Removed. The standalone "Local" provider is no longer emitted when 2+ cookies are present. Each profile's web data stands on its own.

**4. Security — cookie exposure** ✅
`opencode:getProfiles` now strips cookie values before returning. The renderer only receives `{ enabled }` per profile.

**5. i18n** ✅
All previously hardcoded Simplified Chinese strings replaced with `t()` calls. Added 17 new i18n keys across en / zh-TW / zh-CN (51 entries total). The native `confirm()` dialog still uses the browser default as replacing it with a custom modal was a larger UI change than seemed appropriate for this pass, but its text now comes from i18n.

**6. Parallel fetching** ✅
Both `fetchOpenCodeLimits` and the `opencode:status` IPC handler now use `Promise.all` instead of serial iteration. The former also uses the parallel `fetchSingleOpenCodeProfile` with individual 15s timeouts.

Additional cleanups:
- `renderOpenCodeAccountGroup` had a hardcoded `"accounts"` string → moved to i18n key `settings.opencode.nAccounts`
- Removed dead `refreshOpencodeStatus` function
- Aligned the opencode toggle handler with the shared `setAccountGroupExpanded` pattern (consistent with Cursor/Codex/DeepSeek panels)

Rebased onto the latest upstream/main to stay current.

Thanks again — the feedback on accountKey stability and the local/web model was particularly helpful.

Javis603:
Merging this. The multi-account core is solid: deriving the accountKey from the workspace id with a cookie-hash fallback, dropping the standalone Local card in multi mode, stripping cookies out of `getProfiles`, and routing the new strings through i18n all came out clean. The data layer is exactly where it should be.

Two notes for what follows, neither blocking the merge:

I'm going to take the OpenCode settings UI from here and align it with the existing account panels myself. I keep a fairly specific look consistent across the app, so that part is on me, not something for you to chase.

One process note for the next ones: the cache hit rate line on the dashboard is really a separate feature from the account work, so each PR lands cleaner and reviews faster if a change like that comes as its own PR rather than riding along. Same idea keeps the upcoming ones focused.

Thanks for the careful passes here. The accountKey and local/web reworks in particular were well done.

PR #26, opened 2026-06-21
feat: progressive loading — push partial results during full scan

xxtt-01:
## Summary

During a full (3-scan) collector tick, push intermediate results to the UI as each period completes, so users see data incrementally instead of waiting for all three scans.

This is split from the original PR #23 per review feedback. The anchor persistence and dir-skip optimizations will follow in a separate PR after rework.

### Changes

**src/shared/collector.js** (2 sites):

1. **collectUsageOnce** (full-scan path only): Fire onProgress after --today and --month scans, each with whatever period data is available so far.

2. **performTick**: onProgress handler constructs a partial summary (without history/limits keys) and calls onUpdate with reason='progress'.

**Key design decisions:**

- **history and limits are omitted (not null)** from the partial summary — this is the fix for review item #3. carryDeviceHistory carries forward the previous values only when the key is absent; setting it to null bypasses the contract and causes the UI to flash empty.

- **Anchored (watch) ticks** do NOT fire onProgress — they only run one --today scan so progressive loading adds no value.

- **Partial data is Windows-only** (WSL merge happens after all three scans). The final onUpdate with reason 'interval' or 'watch' has the fully merged data.

This PR addresses review item #3 from the original PR #23:
"Progressive updates break the carry-forward contract and tend to flash empty. onProgress explicitly passes history: null and limits: null ... Just omit those two keys on the progressive pushes."

The other four items (dir depth, WSL anchor, config binding, tests) will be addressed in a separate PR after rework.

### DIFF
--- a/devlog/src.md
+++ b/devlog/src.md
@@ -1,0 +1,9 @@
+## 2026-06-21 14:00: 实现渐进推送 onProgress 回调
+- **文件:**
+  - `src/shared/collector.js`
+- **原因:** PR #23 审查反馈 — 需要将渐进推送拆分为独立 PR，并修复 null key 问题
+- **决策:**
+  - 在完整扫描路径（3 次 tokscale 调用）中，每次 period 完成后通过 `onProgress` 推送中间结果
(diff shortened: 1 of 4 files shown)


xxtt-01:
撤回，需要先全面深度检查后再提交。

PR #27, opened 2026-06-21
feat: progressive loading — push partial results during full collector scan

xxtt-01:
## Summary

During a full (3-scan) collector tick, push intermediate results to the UI as each period completes. Users see usage data incrementally instead of waiting for all three tokscale scans to finish.

This PR is split from the original #23 per your suggestion. It contains only the progressive loading feature; the anchor persistence and dir-skip optimisations will follow in a separate PR after rework.

### Changes

**`src/shared/collector.js`** (+24/-3 lines)

1. **`collectUsageOnce` (full-scan path only):** Fire `onProgress` after `--today` and `--month` scans. Both calls are wrapped in try-catch so that a throw inside the callback cannot abort the remaining scans.

2. **`performTick`:** `onProgress` handler constructs a partial summary (without `history`/`limits` keys) and forwards it via `onUpdate` with `reason='progress'`, also wrapped in try-catch.

### Key design decisions

- **`history` and `limits` are omitted (not `null`)** from the partial payload. `carryDeviceHistory` only carries forward the previous values when the key is absent; setting them to `null` bypasses that contract and causes the UI to flash empty (the bug from #23).

- **Anchored (watch) ticks** do NOT fire `onProgress` — they only run one `--today` scan, so progressive loading adds no value.

- **Error resilience:** Both the `onProgress` call sites inside `collectUsageOnce` and the `onUpdate` call inside the `performTick` handler are wrapped in try-catch. This prevents a synchronous throw from silently dropping the remaining scans until the next interval tick (up to 5 minutes).

### Tests

`tests/shared/collectorProgressiveLoading.test.js` (3 tests):

| Test | What it covers |
|------|---------------|
| Full scan fires onProgress twice | Payload shape, key omissions (no history/limits) |
| Anchored tick skips onProgress | Guard condition correctness |
| Throw does not abort full scan | Error resilience — all three periods still populated |

Item #3 from PR #23 review:

> "Progressive updates break the carry-forward contract and tend to flash empty. onProgress explicitly passes `history: null` and `limits: null` ... Just omit those two keys on the progressive pushes."

Fixed: both keys are now omitted from the partial payload. The remaining four items (dir depth, WSL anchor, config binding, coverage) will be addressed in a follow-up PR.

### Additional hardening

During self-review, I also identified and fixed a secondary issue: if `onUpdate` throws synchronously inside the `onProgress` callback, `collectUsageOnce` would reject mid-scan and the month/allTime scans would never execute. Both the `onProgress` invocation and the `onUpdate` call are now protected by try-catch.


Javis603:
Splitting this out was the right call, and the parts that hold up:

- Omitting `history` instead of sending `null` is the correct read of the carry-forward contract.
- Keeping anchored (watch) ticks out of the progressive path is right, they are already a single `--today` scan.
- The `try` / `catch` at the `onProgress` sites guards a synchronous throw in the callback.

A few things I ran into, roughly in order of weight:

1. Two existing tests fail on this branch: `collectorLoadGuards` (the "watch event during an in-flight tick" case) and `collectorTodayDelta`. Both assume one `onUpdate` per full tick and assert `calls.length === 3` after the first update. Firing `onProgress` mid-tick makes `onUpdate` fire after the first scan, so those assertions see 1 or 2 spawns instead of 3. The run also hangs without `--test-force-exit`.

2. The heavier one is sync / agent mode. The sync collector and the agent both post to the hub on every `onUpdate` regardless of `reason`. The progressive call is never awaited, and the final call is only awaited when the consumer returns its promise: `startSyncCollector` does, while the agent callback starts `deliver(...).catch(...)` and returns immediately. So a full tick fires the partial and the final ingest concurrently (the surrounding `try` / `catch` only covers a synchronous throw, not the in-flight POST), and since the hub stores last write wins with no freshness check, a slower partial can land last and overwrite the complete record with `month: emptyPeriod()` / `allTime: emptyPeriod()` until the next tick.

3. `limits` doesn't actually carry forward. The omit trick works for `history`, but on the sync path `postToHub` re-adds `limits: syncLimits(undefined)`, and the local path's `carryDeviceHistory` only carries `history`. So the partial pushes blank limits out at the consumer boundary the same way.

The root of these is that `onProgress` reuses the same `onUpdate` channel that posts full device records to the hub. Progressive loading only really buys anything for the local widget on a cold start, the partials shouldn't reach the hub at all. Note this also means a sync-mode widget renders from the hub SSE stream, so a local-only version wouldn't give it progressive loading at all. Done properly it wants to be a local-render-only path that never posts, plus carrying `month` / `allTime` / `limits` forward the way you did for `history` so warm full scans don't flash those cards to zero.

Given that, and that #29 already makes startup fast on same-day restarts, I'm inclined to hold this one for now rather than take on the extra sync-path surface. I'll leave it open in case you want to bring it down to a local-only version, otherwise no problem, the idea is noted.

xxtt-01:
Thank you for the detailed review. All three issues have been addressed:

**1. Tests** — The onProgress handler no longer calls onUpdate. Instead it calls a dedicated onPreview callback. Tests that use startCollector don't pass onPreview, so progressive pushes no longer interfere with their assertions.

**2. Sync/agent data corruption** — onPreview is only wired for startLocalCollector. Sync, host, and agent modes never see partial data, so no partial POSTs reach the hub.

**3. Limits carry-forward** — The local collector's onPreview callback manually carries forward limits from the previous localDevice, since carryDeviceHistory only handles history. The limits panel no longer flashes empty between scans.

The implementation is now a local-render-only path as you suggested — partial results reach the local widget but never the hub.

Javis603 (changes requested):
The redesign is right. `onProgress` now goes through a dedicated `onPreview` that's only wired for the local collector, so sync/host/agent never see partials and the hub-overwrite race is gone. That was the heavy one.

One thing left before this goes in: warm full scans still replace the complete device with an incomplete preview. After the today scan, month and allTime are filled with `emptyPeriod()`. After the month scan, allTime is still empty. Both previews are also emitted before WSL is refreshed, so the previous WSL contribution disappears from the displayed periods until the final update. `clientStatus` is derived from today or month during those previews as well, so historical-only or WSL-only clients can temporarily move out of the active state.

The preview needs to preserve the distinction between "not scanned yet" and "genuinely zero" across the whole partial record:

- Leave month / allTime omitted until their scans complete, and have the local handler carry the previous values only when those keys are omitted.

- Merge the previous frozen `wslAnchor` into the periods included in a warm preview, then replace it with the freshly scanned WSL bundle in the final update.

- Keep the previous `clientStatus` until allTime is available, with an empty or provisional fallback only on a cold start.

History and limits already follow the same carry-forward idea. Once the usage periods, WSL contribution, and client status stay stable during warm scans, I think this is in good shape.

xxtt-01:
Thank you for the detailed review. I've addressed all three concerns in the latest commit (d1bcaf0):

**1. month/allTime no longer filled with `emptyPeriod()` during warm scans**
The `onProgress` handler now omits `month` and `allTime` from the preview until their respective scans complete. The `onPreview` callback in `main.js` captures carry-forward flags before `summaryWithArchivedClientUsage` processes the summary (which internally fills undefined periods via `targetPeriod` → `normalizePeriod(undefined)` → `emptyPeriod()`), then carries forward the previous device's values when omitted.

**2. WSL contribution preserved during warm previews**
The frozen `wslAnchor` from the previous full scan is now merged into the preview periods via `mergePeriods`. This ensures the WSL contribution doesn't temporarily disappear between period scans. The final `onUpdate` replaces it with the freshly scanned WSL bundle from `collectWsl`.

**3. clientStatus stable during warm scans**
`clientStatus` is now only derived when `allTime` is available in the preview. During warm scans, `main.js` carries forward the previous `clientStatus` from the last complete update, so historical-only or WSL-only clients stay in the correct state.

All 28 collector tests and the full suite (633/634, one pre-existing failure in `preferenceDragDom.test.js`) pass.

Javis603:
Merged, thanks for sticking with this through the revisions. The progressive path is exactly the local-render-only shape it needed, and omitting the unscanned periods so the handler carries them forward is the right contract.

I folded one follow-up in myself rather than send it back another round: `preview.today` was still host-only, so on Windows + WSL the today card dropped its WSL contribution during a warm scan the same way month/allTime used to. I merged the frozen `wslAnchor` into today the way your code already does for month, and gated both by the anchor's day/month so a cross-day or cross-month scan doesn't briefly add the previous period's WSL before the final scan corrects it, plus a regression test: [be61ad6](https://github.com/Javis603/token-monitor/commit/be61ad641c019f5573b26323757e74933429ed1a).

Thanks again for the work on this.

PR #28, opened 2026-06-21
feat: anchor persistence — reuse full-scan snapshots across restarts

xxtt-01:
## Summary

Persist the collector's full-scan anchor to disk so that after a restart the first interval tick can reuse the cached month/allTime snapshots instead of running three serial tokscale scans. This saves approximately 2/3 of the startup scan time.

This is the anchor-persistence portion split from the original PR #23. The progressive loading fix was already submitted separately in PR #27.

### Changes

**src/shared/collector.js** (+39 lines)

1. **configFingerprint helper** — Deterministic string from `normalizeClientsCsv(clients)` and `allTimeSince`. Used to invalidate the persisted anchor when the tracked clients or start date change.

2. **Anchor loading on startup** — After `startCollector` initialises, `readJson` loads `collector-anchor.json` from `sharedDataDir`. The anchor is accepted only when its `dateKey` matches today and its `configFingerprint` matches the current settings. On mismatch (different day, changed config, or missing/corrupt file) the anchor stays null and a full scan runs as normal.

3. **Anchor saving after full scan** — When a non-anchored tick completes and `onAnchorComputed` fires, the Windows-only anchor, `wslBundle`, and current fingerprint are written to disk. Errors are silently caught — the in-memory anchor is unaffected, and the next restart falls back to a fresh full scan.

4. **Interval-loop anchor reuse** — `loop()` checks whether the in-memory anchor is valid for today. When it is, `runTick` passes `{ todayOnly: true }` so only `--today` is scanned and `applyPeriodDelta` derives month/allTime from the anchor. On date change or config change the anchor becomes invalid and the next interval tick does a full scan.

| # | Issue | Status |
|---|-------|--------|
| 1 | Dir-skip directory depth | **Not included** — per review suggestion ("just keep the persisted anchor and drop the skip entirely") |
| 2 | WSL anchor isolation | **Already correct in upstream/main** — Windows-only anchor + separate wslAnchor |
| 3 | Progressive loading null keys | PR #27 (already submitted) |
| 4 | Anchor not tied to config | **Fixed** — configFingerprint invalidates anchor on any settings change |
| 5 | Tests | **Added** — 3 tests: configFingerprint correctness, edge cases, anchored tick delta derivation |

### Tests

`tests/shared/collectorAnchorPersistence.test.js` (3 tests):

| Test | What it covers |
|------|---------------|
| configFingerprint normalises clients | Whitespace case, different clients, different allTimeSince |
| configFingerprint handles undefined/empty | Edge cases for both inputs |
| Anchored tick with valid anchor | Single --today scan, correct applyPeriodDelta derivation |

### DIFF
--- a/devlog/src.md
+++ b/devlog/src.md
@@ -1,0 +1,16 @@
+## 2026-06-21: Anchor 持久化 — 跨重启复用全量扫描快照
+
+- **文件:** `src/shared/collector.js`
+- **原因:** PR #23 审查反馈 #1/#4 — 启动时复用 anchor 避免全量 month/allTime 扫描，配置绑定防止使用过期 baseline
+- **决策:**
+  - Anchor 持久化到 `sharedDataDir()/collector-anchor.json`，全量扫描后写入，启动时读取
+  - 配置绑定通过 `configFingerprint(clients, allTimeSince)` 实现，配置变化时 anchor 自动失效
+  - Interval loop 在 anchor 有效时使用 `todayOnly`，避免每次都做 3 次全量扫描
+  - Dir-skip 不做（作者建议 "drop the skip entirely"）
+  - WSL 部分不动（上游 main 已经正确）
+- **影响范围:** `src/shared/collector.js` — `configFingerprint` 新函数，`startCollector` 中 anchor 读写，`loop` 的 todayOnly 优化
+- **安全分析:**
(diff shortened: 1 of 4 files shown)


xxtt-01:
撤回，等待作者审查改动后再提交。

PR #29, opened 2026-06-21
feat: anchor persistence — reuse full-scan snapshots across restarts

xxtt-01:
## Summary

Persist the collector's full-scan anchor to disk so that after a restart the first interval tick can reuse cached month/allTime snapshots, saving approximately 2/3 of the startup scan time. A 1-hour safety net ensures the anchor is periodically refreshed even in long-running sessions.

This is the anchor-persistence portion split from the original PR #23. The progressive loading fix was already submitted separately in PR #27.

### Changes

**src/shared/collector.js** (+53/-1 lines)

1. **configFingerprint helper** — A deterministic string from `normalizeClientsCsv(clients)` and `allTimeSince`. Used to invalidate the persisted anchor when tracked clients or the start date change. Placed before `startCollector` as a module-level pure function.

2. **Anchor loading on startup** — After `startCollector` initialises its state, `readJson` loads `collector-anchor.json` from `sharedDataDir`. The anchor is accepted only when both `dateKey` matches today and `configFingerprint` matches the current settings. On mismatch (different day, changed config, or missing/corrupt file) the anchor stays null and the first interval tick runs a full scan. `lastFullScanAt` is set to `Date.now()` on load so the 1-hour timer starts from app launch.

3. **Anchor saving after full scan** — When a non-anchored tick completes, the Windows-only anchor, `wslBundle`, and current fingerprint are written to the anchor file. Errors are silently caught — the in-memory anchor is unaffected, and the next restart falls back to a fresh full scan.

4. **Interval-loop optimisation** — `loop()` checks two conditions before each tick: whether a full scan is due (1-hour timer) and whether the in-memory anchor is valid for today. When the anchor is valid and no full scan is due, the tick runs with `todayOnly: true`, scanning only `--today` and deriving month/allTime via `applyPeriodDelta`. This is mathematically exact — the delta formula `month = anchor.month + (fresh_today - anchor_today)` has zero accumulated error because it always computes against the anchor baseline, not the previous tick's result.

### Safety analysis

Three independent protection layers ensure correctness:

| Layer | Mechanism | Trigger |
|-------|-----------|---------|
| Date key | `anchor.dateKey !== localTodayKey()` | Midnight or timezone change |
| Config fingerprint | `configFingerprint(clients, allTimeSince)` mismatch | Settings change (restarts collector) |
| Periodic full scan | `Date.now() - lastFullScanAt >= FULL_SCAN_INTERVAL_MS` | 1 hour since last full scan |

Any one of these triggers a full scan, which updates the anchor and resets `lastFullScanAt`.

Edge cases verified during review:

- **First run (no anchor file):** `readJson` returns null → full scan → anchor saved
- **Same-day restart:** Anchor loaded with matching dateKey and fingerprint → todayOnly ticks
- **Cross-day restart:** dateKey mismatch → full scan → new anchor with today's date
- **Config change at runtime:** `startMode()` restarts collector → new `startCollector` → fingerprint mismatch → full scan
- **File write failure:** try-catch swallows → in-memory anchor stays valid → next restart self-heals
- **Crash during write:** Partial JSON → `readJson` returns null → full scan (self-healing)
- **Old-version upgrade:** No `configFingerprint` field → `undefined !== string` → full scan
- **WSL data:** `wslBundle` persisted alongside anchor → restored on load → `normalizePeriod` validates shape

### Tests

`tests/shared/collectorAnchorPersistence.test.js` (3 tests):

| Test | Coverage |
|------|----------|
| configFingerprint normalises clients | Whitespace/case normalisation, different inputs produce different outputs |
| configFingerprint handles undefined/empty | Edge cases for both parameters |
| Anchored tick with valid anchor | Single `--today` scan, correct `applyPeriodDelta` derivation |

| # | Issue | Status |
|---|-------|--------|
| 1 | Dir-skip directory depth | **Not included** — per review suggestion ("just keep the persisted anchor and drop the skip entirely") |
| 2 | WSL anchor isolation | **Already correct in upstream** — Windows-only anchor + separate wslAnchor |
| 3 | Progressive loading null keys | PR #27 (already submitted) |
| 4 | Anchor not tied to config | **Fixed** — configFingerprint invalidates anchor on settings change |
| 5 | Tests | **Added** — 3 tests covering fingerprint and anchored-tick behaviour |


Javis603:
This direction is good and the implementation is mostly there. What holds up:

- `configFingerprint` invalidating the anchor when the client list or `allTimeSince` changes is the right guard.
- Deriving month / allTime against the persisted anchor is the same exact identity the watch ticks already use, so they stay correct between full scans on append-only logs.
- Falling back to a full scan on a missing or unparseable file, and on an old anchor with no fingerprint field.

One correctness item I'd want changed before it goes in: on load you set `lastFullScanAt = Date.now()`, so the hourly safety-net scan is measured from launch, not from when the anchor was actually taken. A session that restarts more often than once an hour can keep pushing that full scan out indefinitely. The delta stays exact for plain append-only usage, so that scan is mostly a safety net, but a few things still depend on it actually running: the WSL snapshot only refreshes on a full tick, and re-priced history or a same-day import / delete of older data is only reconciled by one. Persisting the real `fullScanAt` and deciding from its actual age on load would close that.

Smaller items:

- The persistence test covers `configFingerprint` and the anchored tick, but not the actual disk read/write, the cross-day / expiry invalidation, or the restart reuse, which is the part most worth pinning down.
- `devlog/src.md` and `devlog/tests.md` look like local notes, probably don't want those in the repo.
- The new test file isn't in the `check` list.

Once the `lastFullScanAt` part is sorted I think this is in good shape.

xxtt-01:
Thank you for the thorough review. All four items have been addressed:

1. fullScanAt persistence: The hourly safety-net timer is now driven by the anchors actual fullScanAt timestamp rather than Date.now() on load. Old anchors without the field fall back to lastFullScanAt = 0 which self-heals once a new anchor is saved.

2. Test coverage: Two new integration tests added - restart reuse (writes an anchor, starts collector, verifies todayOnly) and cross-day invalidation (stale dateKey triggers full scan).

3. Test file registration: collectorAnchorPersistence.test.js added to package.json check list.

4. Devlog files: Removed from git tracking (git rm --cached + .gitignore).

All 38 collector tests pass.

Javis603 (changes requested):
fullScanAt persistence is the right fix, and driving the hourly scan off the anchor's real timestamp is exactly it. The shared-dir isolation you added to the existing collector tests is a good catch too.

Three things before merge:

- Rebase onto current main and drop the package.json change entirely. main replaced the `check` script with a `node --test "tests/**/*.test.js"` glob, so the new test file is picked up automatically and doesn't need registering. That `check` block is also the only merge conflict right now, so dropping it clears the conflict and lets CI run.

- The safety-net scheduling trusts the timestamp without validating it. A missing, unparseable, or future fullScanAt leaves lastFullScanAt at 0 (or in the future), and since the due-check needs `> 0` plus an elapsed hour, the hourly full-scan safety net gets skipped until the anchor is rebuilt by a day rollover, a config change, or a manual full scan (a future timestamp defers it to that time plus an hour). Since this code always writes a valid timestamp the realistic trigger is clock skew, but it quietly defeats the age-driven rescan. Treating an untrustworthy timestamp as "force a full scan" (or invalidating the anchor) plus tests covering missing, invalid, and future timestamps would close it.

- One more on WSL, and this is a requirement I missed in my earlier review. WSL has no file watcher, so the 5-minute interval scan is its only near-real-time refresh path. My earlier note treated the hourly full scan as WSL's refresh, but anchoring the interval tick drops WSL from ~5 minutes to ~1 hour: anchored interval ticks now reuse `wslAnchor`, so the default 5-minute tick refreshes the Windows host's `today` but doesn't scan WSL at all, and new WSL usage stays stale until the next full scan, a manual refresh, or an anchor invalidation. The host-side optimization should stay; the piece to add is that interval ticks still refresh WSL while watch-triggered ticks keep reusing the frozen `wslAnchor` (rescanning WSL on every few-second watch tick would be too heavy). A test that WSL is rescanned on interval ticks but not on ordinary watch events would pin it.

Once those three are sorted I think this is in good shape.

xxtt-01:
Thank you for the thorough second review. All three issues have been addressed in the latest commit (a1a4a62):

**1. Rebase onto current main + drop package.json change** ✅
Rebased onto `upstream/main` (0295fff). The `check` script additions have been removed from `package.json` — `node --test "tests/**/*.test.js"` auto-discovers the new test files.

**2. Safety-net timestamp validation** ✅
The anchor loading code now only accepts `fullScanAt` timestamps that are both valid (`Number.isFinite`) and not in the future (`parsed <= Date.now()`). Invalid, unparseable, future, or missing timestamps leave `lastFullScanAt` at 0, and `loop()` now treats `lastFullScanAt === 0` as a forced full scan (previously it silently skipped the safety net).

Two new tests cover this:
- `future fullScanAt forces a full scan on first interval tick` — writes an anchor with a future timestamp, verifies 3 spawns
- `missing fullScanAt forces a full scan on first interval tick` — writes an anchor without `fullScanAt`, verifies 3 spawns

**3. WSL refresh on interval ticks vs watch ticks** ✅
Added a `refreshWsl` option to `collectUsageOnce` that decouples WSL scanning from the anchor optimization. Three distinct paths now exist:
- **Interval anchored ticks** (`refreshWsl: true`): scan WSL fresh every 5 minutes while keeping the tokscale anchor optimization (1 spawn). The frozen `wslAnchor` is updated after each interval refresh so subsequent watch ticks see current WSL data.
- **Watch ticks** (no `refreshWsl`): reuse the frozen `wslAnchor` as before — WSL is too heavy for sub-second watch cycles.
- **Full (non-anchored) ticks**: unchanged — WSL is scanned as part of the complete rescan via `!anchorUsed`.

A new test in `collectorWslMerge.test.js` verifies that `refreshWsl: true` rescans WSL and produces fresh data, while a subsequent call without `refreshWsl` reuses the anchor.

All 29 collector tests and the full suite (653/654, one pre-existing failure in `preferenceDragDom.test.js`) pass on the rebased branch.

Javis603:
Landed in main, thanks for working through all the rounds on this. The anchor persistence is a clean win: reusing the full-scan snapshot across restarts so the first interval tick only needs `--today`, with the `fullScanAt` validation (invalid / future / missing all force a full scan) and the WSL refresh decoupled from the anchor so interval ticks keep WSL fresh while watch ticks stay frozen. All three points from the last round check out.

Because #27 (the progressive-preview work) landed after you rebased, your branch conflicted with current main in `collector.js`, so I rebased onto main and squashed it in, resolving the conflict to keep both the #27 preview path and your anchor/WSL logic: [f0cfc8f](https://github.com/Javis603/token-monitor/commit/f0cfc8f). GitHub can't show this as "merged" since the squash changes the commit SHA, but the commit keeps you as author.

One small follow-up from me on top: a couple of the collector tests (including one I added in #27) weren't isolating `TOKEN_MONITOR_SHARED_DIR`, so once your on-disk anchor persistence landed they could write a real anchor into the user data dir and make the suite order-dependent. I isolated those and added an explicit unparseable-`fullScanAt` case: [f555720](https://github.com/Javis603/token-monitor/commit/f555720). Full suite is green. Thanks again.

PR #32, opened 2026-06-22
feat(limits): add Grok and Minimax quota providers

SDS-CC:
# feat(limits): add Grok and Minimax quota providers

## Summary

Two new **AI Tool Limits** providers, plus a small scaffold (`hashKey`) shared
by both.

| Provider  | Source                                                | Window       |
|-----------|-------------------------------------------------------|--------------|
| Grok      | `cli-chat-proxy.grok.com/v1/billing`                  | Monthly      |
| Minimax   | `api.minimax.io/v1/api/openplatform/coding_plan/remains` (auto-retry CN) | 5h + Weekly |

The Grok rewrite drops the previous hand-written gRPC-Web + protobuf decoder
in favor of the JSON+HTTP endpoint TokenTracker uses locally. The Minimax
fetcher is new. Both go through the existing `collectLimitsOnce` →
`normalizeLimitProvider` pipeline; no API surface changes outside the two
new modules.

### 1. Grok — rewrite to plain JSON+HTTP

`src/shared/grokLimits.js` (replaces the prior gRPC-Web decoder).

- Endpoint: `cli-chat-proxy.grok.com/v1/billing` — same one TokenTracker's
  `src/lib/grok-limits.js` calls. Response shape:
  `{ config: { monthlyLimit, used, billingPeriodEnd } }` → one monthly
  window with `usedPercent = used / monthlyLimit * 100`.
- Auth precedence: explicit `grokBearerToken` setting →
  `$GROK_BEARER_TOKEN` env → `~/.grok/auth.json` (OIDC scope preferred,
  legacy `/sign-in` fallback, then any non-empty entry). The cookie-paste
  path that was in the previous Grok fetcher is intentionally dropped —
  TokenTracker doesn't ship one either.
- On-demand cap (the second `config` field pair) is intentionally NOT
  surfaced; the existing per-provider column has only one slot for Grok.

### 2. Minimax — new provider

`src/shared/minimaxLimits.js`.

- Endpoint: `api.minimax.io/v1/api/openplatform/coding_plan/remains` —
  global endpoint by default; retries `api.minimaxi.com` (CN) automatically
  when the token is rejected (401/403), the same pattern CodexBar uses.
  Pass `minimaxApiHost: 'cn'` / `'en'` / `'minimax.io'` to pin a single
  region (still supported).
- Reads `data.model_remains[]` (the live response nests the array under
  `data`; we also accept the top-level fallback for older proxies /
  fixtures). Picks the row where `model_name === 'general'`; ignores
  `video`, `voice`, etc.
- Windows: 5h session (`current_interval_remaining_percent` → `end_time`)
  and weekly (`current_weekly_remaining_percent` → `weekly_end_time`).
  Percents can arrive as strings; both are clamped to `[0, 100]`.
- Only suppresses a row when `*_status === 3 && percent is null or ≥100` —
  the placeholder lane CodexBar / cc-switch both guard against. Live
  responses don't always carry the status field, so we don't gate on it
  being present.
- Auth: Bearer from explicit `minimaxApiKey` setting →
  `$MINIMAX_TOKEN_PLAN_KEY` → `$MINIMAX_API_KEY`.
- The resolved region (`'en'` / `'cn'`) flows up through
  `normalizeLimitProvider(region)` so the renderer's "Open Token Plan
  page" button routes to the matching `platform.minimax.io` /
  `platform.minimaxi.com` host. Both are now in the
  `isAllowedExternalUrl` allowlist.

### 3. Shared scaffold

- `src/shared/hashKey.js` — sha256-based `hashKey(...parts)` for stable
  `accountKey` derivation. Extracted from the inline copies that lived in
  `grokLimits` / `minimaxLimits` so the two providers don't form a require
  cycle with `limitCollector` (which already imports them).
- `limitCollector.js` — registers `grok` and `minimax` providers,
  passes `grokCredential` / `minimaxApiKey` from settings into the
  per-provider fetch options.
- `limits.js` — adds `'minimax'` and `'grok'` to `VALID_PROVIDERS`; adds
  `region` to the normalized provider result.

### 4. UI / i18n / renderer wiring

- New accordion blocks for Grok and Minimax in the limits settings panel
  (`src/electron/renderer/index.html` + `app.js`)
- i18n keys for `en`, `zh-CN`, `zh-TW` (`src/electron/renderer/i18n.js`)
- Status presentation for the two new providers in
  `limitProviderPresentation.js` (Add API key / Run grok login /
  Subscription tag, etc.)
- README updates so the supported-tools table and limit-providers
  default list reflect the new providers
- `.env.example` entries for `MINIMAX_TOKEN_PLAN_KEY` and
  `GROK_BEARER_TOKEN`

## What's NOT in here (moved to a separate PR)

Reviewer asked for the following to be split out into their own PR:

- **`safeStdio`** (`installSafeStdout` + `safeLog`/`safeWarn` wrappers)
  — fixes the "JavaScript error in the main process" dialog from
  background log EPIPE. Independent of the limits feature.
- **`applyProxyFromEnv` / `TOKEN_MONITOR_PROXY`** — routes every Node
  `fetch()` through an env-configured HTTP/SOCKS proxy. The Grok
  endpoint is unreachable in some regions without it, but the change
  is app-wide network behavior and belongs in its own PR.

Both live on the `feat/safe-stdio-proxy` branch in this fork and will
be opened as a separate PR against upstream.

## Second-round fixes (post-review)

Follow-up items from a second review pass against the cleaned-up PR:

- `grokLimits.js` — removed a local Windows machine path from the
  comment that pointed at `C:\Users\Administrator\AppData\Local\Programs\…`
  (a TokenTracker install on the contributor's box). Replaced with a
  generic "TokenTracker's `src/lib/grok-limits.js`" reference.
- `main.js` — `currentGrokCredential` had a dead `readAuthJson(env)`
  fallback after `grokCredential(env)` (which already falls back to
  `~/.grok/auth.json` internally, so the second call is unreachable).
  Collapsed to `return grokCredential(process.env);` and dropped the
  now-unused `readAuthJson` import.
- `minimaxLimits.js` — renamed `toMillisNumber` → `parseNumberOrNull`.
  The old name suggested "milliseconds" but the function is a generic
  "parse a number-or-null numeric value" used for status codes, percent
  strings, and timestamps. Updated the comment to reflect that the unit
  is up to the caller.
- `minimaxLimits.test.js` — added a test documenting what happens when
  the live response returns a `current_weekly_status: 3` lane with a
  real (non-100, non-null) percent: we currently render it as a real
  window, matching the CodexBar placeholder guard. If the live
  endpoint ever starts returning that shape, the percent is probably
  a stale placeholder and this test will need to flip to assert the
  lane is suppressed.

## Files

```
.env.example                                       |  16 +-
README.md                                          |  10 +-
README.zh-CN.md                                    |   5 +-
README.zh-TW.md                                    |   5 +-
src/electron/main.js                               |  99 +++---
src/electron/renderer/app.js                       | 165 +++++++++
src/electron/renderer/i18n.js                      |  57 ++++
src/electron/renderer/index.html                   |  43 +++
src/electron/renderer/limitProviderPresentation.js |  17 +-
src/shared/grokLimits.js                           | 223 ++++++++++++
src/shared/hashKey.js                              |  11 +
src/shared/limitCollector.js                       |  50 ++-
src/shared/limits.js                               |  12 +-
src/shared/minimaxLimits.js                        | 268 +++++++++++++++
tests/electron/rendererClientLabels.test.js        |   2 +-
tests/shared/deepseekLimits.test.js                |   5 +-
tests/shared/grokLimits.test.js                    | 265 +++++++++++++++
tests/shared/hashKey.test.js                       |  40 +++
tests/shared/minimaxLimits.test.js                 | 374 +++++++++++++++++++++
```

## Tests

- `tests/shared/grokLimits.test.js` — JSON parsing, `auth.json`
  resolution (OIDC preferred, legacy fallback, any-key fallback,
  missing file, invalid JSON), `fetchGrokLimits` for 200 / 401 / 500 /
  network-error, credential precedence (settings > env > auth.json)
- `tests/shared/minimaxLimits.test.js` — token normalization
  (whitespace + quote stripping), `data.model_remains` nested shape
  (live response), top-level fallback, status==3 placeholder
  suppression, percent-as-string parsing, clamping, second-precision
  timestamps, region retry on 401, no-retry on 5xx, `parseLimitProviders`
  default set now includes both
- `tests/shared/hashKey.test.js` — digest format, stability, the
  `part || ''` guard for undefined/null/''
- `tests/shared/deepseekLimits.test.js` — provider list update

## Validation

- `npm run lint` → 10 pre-existing errors in `main.js` /
  `limitCollector.js` (`isHubConfigured`, `currentOpenCodeCookie`,
  `readCodexAuthIdentity`, unused `options` args, etc.). All already
  broken on `upstream/main` before this PR; no new lint issues.
- `npm test` → 700/701 pass. The 1 failure is in
  `preferenceDragDom.test.js` and is pre-existing on `upstream/main`
  (`main section holds views; appearance is its own section; window
  holds behavior and presence`). Reviewer explicitly noted that one
  is theirs to fix separately.

## Security

- The Grok rewrite reads the bearer token from `~/.grok/auth.json` but
  only transmits it to `cli-chat-proxy.grok.com` over HTTPS.
- The cookie-paste path is gone — no browser cookie sniffing.
- `platform.minimax.io` and `platform.minimaxi.com` are both in the
  external-URL allowlist, but the URLs are hardcoded in the click
  handler, not interpolated from user input.


SDS-CC:
Pushed one more commit (`5c3dd95`) that addresses a circular dependency introduced by the previous PR cleanup.

The cleanup commit re-pointed `grokLimits.js` and `minimaxLimits.js` at `limitCollector`'s `hashKey`, but `limitCollector` already `require`'s those two providers for their `grokCredential`, `readAuthJson`, `parseGrokBilling`, and `fetchGrokLimits` exports. The resulting cycle made `hashKey` `undefined` at module load time:

```
Warning: Accessing non-existent property 'hashKey' of module exports inside circular dependency
(…same for grokCredential / readAuthJson / parseGrokBilling / fetchGrokLimits)
```

The first `fetchGrokLimits()` call saw `hashKey === undefined` and threw inside the `accountKey` assignment, the catch wrapper returned `status: 'unavailable'`, and the widget's Grok row went blank even though the HTTP request itself still succeeded.

Fix: extract `hashKey()` to `src/shared/hashKey.js`, a leaf module that depends only on `node:crypto`. `limitCollector` now requires it (and drops its local copy), and `grokLimits` / `minimaxLimits` require it directly. The cycle disappears.

Tests in `tests/shared/hashKey.test.js` cover the digest format (`sha256:` prefix + 64 hex chars), stability across calls, the `part || ''` guard that maps `undefined` / `null` / `''` to the same digest, and the prefix that downstream code parses.

Javis603:
Thanks for this, the Grok rewrite in particular is a good call.

## Grok

Moving off the gRPC-web + protobuf path to `cli-chat-proxy.grok.com/v1/billing` is the right tradeoff for a widget that just needs the quota and degrades gracefully. I cross-checked the response shape against TokenTracker's `src/lib/grok-limits.js` (it is open source): same endpoint, same `{ config: { monthlyLimit, used, billingPeriodEnd } }`, same `used / monthlyLimit * 100`. So the shape is verified and I'm happy with this part.

## Minimax (needs a fix before merge)

The parser doesn't match the live `coding_plan/remains` response. Two issues, both verified by running the PR's own `parseMinimaxTiers` against a real Token Plan response:

1. Shape: the real response nests the array under `data` (`data.model_remains`), but `selectMinimaxGeneralBucket` reads top-level `body.model_remains`, so it finds nothing.
2. Status gate: `buildMinimaxSessionWindow` / `buildMinimaxWeeklyWindow` require `current_interval_status === 1` / `current_weekly_status === 1`, but the live response doesn't include those fields at all.

Net effect against a real response:

```
parseMinimaxTiers(realResponse)       -> []   (status: unavailable)
parseMinimaxTiers(realResponse.data)  -> []   (still empty, no status fields)
// windows only appear once status:1 is injected by hand
```

The test fixtures use a flattened shape with explicit `status: 1`, which is why the suite stays green, but that isn't what the endpoint returns. A real response looks like:

```json
{
  "base_resp": { "status_code": 0, "status_msg": "success" },
  "data": {
    "current_subscribe_title": "Token Plan Plus",
    "model_remains": [
      {
        "model_name": "general",
        "current_interval_remaining_percent": "96",
        "start_time": 1780279200000,
        "end_time": 1780297200000,
        "current_weekly_remaining_percent": "99",
        "weekly_start_time": 1780243200000,
        "weekly_end_time": 1780848000000
      }
    ]
  }
}
```

There is no `current_interval_status` / `current_weekly_status`, and the percents can be strings. I cross-checked the shape against CodexBar's MiniMax fetcher, which reads `data.model_remains` with a top-level fallback and doesn't gate on a status field. It only hides the `status == 3 && count == 0 && remaining == 0 && pct >= 100` placeholder lane.

Suggested fix:

- read `body.data.model_remains` first, fall back to top-level
- drop the hard `status === 1` gate; emit the window whenever a percent is present, and only suppress the status==3 placeholder lane
- rebuild the fixtures from the real shape so the tests guard the right contract

One more, separate from the parser: region selection defaults to the CN host (`api.minimaxi.com`) and only reaches the global host (`api.minimax.io`) via `minimaxApiHost`, which `main.js` never forwards from settings. So a global (minimax.io) account can't be fetched at all and has no way to switch. Rather than wiring a manual toggle, it's cleaner to try both automatically: default to global and retry CN when the token is rejected, the way CodexBar does it, so the right region is picked without the user knowing about it. The "Open Token Plan page" button has the same problem: it's hardcoded to `platform.minimaxi.com` and doesn't redirect, so a global account lands on the wrong page. It should follow the same region (`platform.minimax.io` vs `platform.minimaxi.com`) and the external-URL allowlist should accept both hosts.

Two ways forward, your call: you push the fix, or I can push the Minimax fix and you keep the Grok + scaffolding credit.

## Please split safeStdio out

The EPIPE fix looks reasonable on its own, but it's unrelated to the limit providers. Could you move it to its own PR? That keeps the providers reviewable independently from the stdout change.

## Drop the global fetch proxy

The PR also adds a global fetch proxy (`applyProxyFromEnv` / `TOKEN_MONITOR_PROXY`) that routes every outbound `fetch()` through `setGlobalDispatcher`, plus a `TOKEN_MONITOR_NO_PROXY` in `.env.example` that nothing reads. It isn't part of the limits feature and it changes app-wide network behavior, so I'd rather not take it in here. Could you drop it from this PR?

## Minor (not blocking)

- The PR description lists `current_window_remaining_quota` / `weekly_remaining_quota` / "status 0 or 1", but the code reads `current_interval_remaining_percent` / `current_weekly_remaining_percent` and gates on `=== 1`. Worth aligning the two.
- The `cc-switch-reference/coding_plan.rs` path mentioned in the comment isn't in the PR.

Don't worry about the 22 lint errors or the 2 `preferenceDragDom` failures, those are pre-existing on main and not yours to fix. I'll handle them separately.

Once Minimax is sorted and the safeStdio and proxy changes are out of this PR, I think it's in good shape.

Javis603:
Thanks, the revision covers the review: the Minimax parser now reads the nested `data.model_remains`, drops the hard status gate, auto-falls-back between the global and CN hosts, and the Open Token Plan button follows the resolved region. safeStdio is split out into #35 and the proxy is gone. Nice work.

Two housekeeping things I handled on top so this lands cleanly:

1. The branch was behind `main` and conflicting, so I pushed a merge commit that brings it onto current `main`. Your provider logic (`grokLimits.js`, `minimaxLimits.js`, `hashKey.js` and their tests) is untouched.
2. The branch had drifted from `main` on a couple of unrelated files: the `onPreview` progressive-push block in `main.js` had been dropped, and `.env.example` plus the READMEs had lost the `pi,zed,kilocode` entries. I restored those from `main` and added the Grok/Minimax rows to all three READMEs.

Verified locally: `npm run lint` clean, `npm test` 723/723.

One follow-up I will do as a separate maintainer commit (not on this PR): the Grok and Minimax `fetch` calls do not carry the 12s abort timeout the other providers use, so a hung endpoint could stall the collector. Small fix, keeping it off your PR.

Merging via squash now. Thanks again!

PR #35, opened 2026-06-23
feat: add safe stdout wrapper to suppress EPIPE dialogs

SDS-CC:
# feat: add safe stdout wrapper to suppress EPIPE dialogs

Standalone fix for an Electron main-process issue: background log traffic
from the collector / sync / hub watchers raises unhandled `EPIPE: broken
pipe, write` rejections when the parent shell closes its end of the
stdout pipe (npm detached, terminal closed, log redirected to a
non-seekable consumer). Electron turns those into a **"JavaScript error
in the main process"** dialog, even though the app itself is healthy.

## What's in here

`src/shared/safeStdio.js` exposes:

- `safeLog(...args)` / `safeWarn(...args)` — write directly to
  `process.stdout` / `process.stderr`, swallow `EPIPE`, re-throw any
  other error so genuine bugs surface.
- `installSafeStdout()` — register no-op `EPIPE` handlers on the
  stream `'error'` event so the **async** EPIPE path is also
  suppressed. Synchronous write failures are caught by `safeLog` /
  `safeWarn`, but backpressure closes raise the error event on the
  next tick.

`installSafeStdout()` runs at startup, before any logger can fire. All
`console.log` / `console.warn` call sites in `src/electron/main.js`
route through `safeLog` / `safeWarn`.

`tests/shared/safeStdio.test.js` covers:
- `safeLog` / `safeWarn` arg coercion (non-string args join with spaces)
- Sync EPIPE suppression
- Re-throw of non-EPIPE errors so genuine bugs surface
- Idempotency of `installSafeStdout`

## Files

```
src/electron/main.js           |  52 +++++++++++----------
src/shared/safeStdio.js        |  45 +++
tests/shared/safeStdio.test.js | 100 +++++++++++++++++++++++++++++++++++++
```

## Origin

Split out from a larger limits-providers PR
([#32](https://github.com/Javis603/token-monitor/pull/32)) per the
maintainer's request. Originally that PR also proposed routing Node
`fetch()` through a `TOKEN_MONITOR_PROXY` env var; that's been dropped
from this branch at the maintainer's direction (the proxy change is
app-wide network behavior and not appropriate for a focused
providers PR).

### DIFF
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -6,12 +6,12 @@ const os = require('node:os');
 const path = require('node:path');
 const { app, BrowserWindow, globalShortcut, ipcMain, nativeImage, screen, session, shell } = require('electron');
 const { defaultDeviceId, generateHubSecret, lanIpv4Addresses, loadDotEnv, pidFilePath, sharedDataDir } = require('../shared/config');
-const { safeLog, safeWarn, installSafeStdout } = require('../shared/safeStdio');
+const { installSafeStdout } = require('../shared/safeStdio');
 const { appVersion } = require('../shared/appVersion');
 
 // Install EPIPE suppression before anything that might log. Without this,
-// a closed parent pipe turns the next log call into an unhandled rejection
-// and Electron pops a "JavaScript error in the main process" dialog.
+// a closed parent pipe turns the next log call into an unhandled 'error'
+// event and Electron pops a "JavaScript error in the main process" dialog.
 installSafeStdout();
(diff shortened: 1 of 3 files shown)


Javis603:
Thanks for splitting this out of #32.

I pushed one commit (2450c30) before merging that trims it to just `installSafeStdout()`: the global `'error'` handler on stdout/stderr is what actually suppresses the dialog (EPIPE surfaces async on the stream, so one handler covers every logger in the main process), which makes the `safeLog` / `safeWarn` wrappers redundant, so those call sites go back to native `console.*`. Kept it Electron-only since the dialog is main-process specific.

PR #40, opened 2026-06-28
Keep ingestion and Copilot detection working after update

a8lo:
## Summary
- restore the hub ingest body-size patch with a configurable 16MB default
- add a Copilot CLI session-state fallback collector for token totals
- watch both Copilot otel and session-state roots, with regression tests
- remove the unused electron-icon-builder dev dependency from the refreshed install

## Verification
- node --test tests/hub/server.test.js tests/shared/copilotCliUsage.test.js tests/shared/collectorLoadGuards.test.js tests/shared/clientStatus.test.js
- npm run lint
- node --check targeted changed JS files
- live hub stats confirmed ubuntu reports copilot active with nonzero usage

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,22 +22,11 @@
         "@eslint/js": "^10.0.1",
         "electron": "42.2.0",
         "electron-builder": "^26.8.1",
-        "electron-icon-builder": "^2.0.1",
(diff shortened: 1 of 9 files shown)


a8lo:
Closing at request of the requester; this change should not be merged.

PR #42, opened 2026-07-01
fix(usage): show Grok token breakdown using explicit input totals

SDS-CC:
## Summary

Grok sessions from tokscale arrive as **input-only** rows: `output` and `cacheRead` are zero, but `input` carries the real usage. The dashboard accordion previously inferred cache-miss as `total - cacheRead - output`, which collapses to zero for Grok and makes expanded rows look empty or misleading.

This PR:

- Aggregates tokscale `input` into `inputTokens`, `clientInputs`, and `modelInputs` in `usage.js`
- Passes explicit input through client/model rows in the renderer
- Extracts accordion math/rendering into `tokenBreakdown.js`
- Keeps the standard three-row layout (cache hit % / cache miss % / output) for Grok models

## Related upstream work

The **under-counted Grok totals** (e.g. ~4M vs ~20M) are fixed separately in tokscale:

- Issue: https://github.com/junhoyeo/tokscale/issues/797
- PR: https://github.com/junhoyeo/tokscale/pull/798

This token-monitor PR is the **display companion**: once tokscale #798 ships, bumping the bundled tokscale version will give correct totals; this PR ensures the expanded breakdown renders correctly for Grok's input-only shape.

## Test plan

- [x] `node --test tests/shared/usage.test.js tests/electron/tokenBreakdown.test.js`
- [x] Manual: expand `grok-composer-2.5-fast` / `grok-build` rows — expect 0% / 100% / 0 with non-zero cache-miss counts when input is present


Javis603:
Thanks @CYCYC-SDS for digging into the Grok breakdown, and for separating the display change from the upstream total under-count so clearly.

Coming back to this after a while: the root cause was the Grok total under-count in tokscale (junhoyeo/tokscale#798), and the bundled tokscale has since moved to ^4.4.1, so that fix is already in. With the totals correct, the existing accordion math (`cacheMiss = total - cacheRead - output`) already renders Grok as 0% hit / 100% miss / 0 output, the same as every other provider, so the explicit-input path isn't needed for the Grok case. `src/shared/usage.js` has also been reworked since (it now carries `inputTokens` through a component-based token model), so this branch no longer applies cleanly.

I'm closing this as resolved by the upstream fix. Thanks again for the thorough write-up.

PR #43, opened 2026-07-01
fix(collector): include Hermes profile databases in tokscale scans

SDS-CC:
## Summary

Hermes Agent stores each profile in its own SQLite database under `$HERMES_HOME/profiles/*/state.db`. Tokscale only scans the default `state.db`, so Token Monitor under-counts Hermes when users run multiple profiles.

This PR teaches the collector to:

- Resolve the real Hermes home (`HERMES_HOME`, Windows `%LOCALAPPDATA%\hermes`, or `~/.hermes`)
- Discover every profile directory that contains a `state.db`
- Inject those paths into `TOKSCALE_EXTRA_DIRS` whenever tokscale is spawned with `--client … hermes …`
- Watch profile directories for filesystem changes so refreshes stay timely

No tokscale fork or manual `settings.json` editing required.

## Evidence

On a machine with one main Hermes DB plus multiple profile DBs, `collectUsageOnce({ clients: 'hermes' })` matches the SQL sum across all discovered profile databases (not just the main `state.db`).

## Test plan

- [x] `npm run lint`
- [x] `node --test tests/shared/hermesProfiles.test.js`
- [x] Manual `collectUsageOnce({ clients: 'hermes' })` matches main + profile SQL totals

### DIFF
--- a/src/shared/collector.js
+++ b/src/shared/collector.js
@@ -12,6 +12,7 @@ const { normalizeClientsCsv } = require('./clientTracking');
 const { tokscalePackageNameForPlatform, tokscalePlatformKey } = require('./tokscalePlatform');
 const { applyPeriodDelta, emptyPeriod, extractUsageFromTokscale, mergePeriods } = require('./usage');
 const { collectWslUsage: collectWslUsageImpl, emptyWslBundle, probeWslState: probeWslStateImpl } = require('./wslUsage');
+const { hermesProfileWatchDirs, resolveHermesHome, tokscaleEnvFromSpawnArgs } = require('./hermesProfiles');
 const { parseGraphResult, normalizeHistory } = require('./history');
 const { collectLimitsOnce, createLimitsCollector } = require('./limitCollector');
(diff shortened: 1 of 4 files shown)


Javis603:
Thanks for this, and for wiring the profile discovery through `TOKSCALE_EXTRA_DIRS` instead of touching anyone's `settings.json`, that's the right call.

I pushed one commit (3c54092) before merging. The profile watch dirs were being silently dropped: the existing Hermes `watchIgnoreMatcher` walks the watch roots parent-first, so it pruned each `profiles/<name>` dir as non-db junk before reaching its own root check. Profile usage still counted on the interval scans, but a live edit to a profile DB never fired a refresh. The commit keeps any explicit watch root first and then prunes, with a regression test covering the profile dir and its `state.db` family. It also reads `LOCALAPPDATA` when resolving the Windows Hermes home, and skips the injection on `--home` (WSL) scans since tokscale ignores env roots there anyway.

PR #51, opened 2026-07-02
chore(deps): upgrade tokscaler for OMP parser fix

GoldTest:
Upgrade tokscaler to include the fix for Oh My Pi (OMP) session parser.

**Root Cause:**  
OMP session files start with a `title` line before the `session` header. The parser was returning empty messages on any non-"`session`" line, causing OMP sessions to be silently skipped.

**Fix:**  
Skip non-"session" lines until the actual session header is found.

**Upstream PR:** https://github.com/junhoyeo/tokscale/pull/805

**Changes:**
- `tokscale`: 4.0.5 → 4.0.8 (includes the parser fix)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "koffi": "^3.0.2",
         "semver": "^7.8.1",
         "tar": "^7.5.15",
-        "tokscale": "^4.0.5"
+        "tokscale": "4.0.8"
       },
       "devDependencies": {
         "@eslint/compat": "^2.1.0",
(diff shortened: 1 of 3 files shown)


Javis603:
Thanks for digging into this, and especially for the parser fix you opened upstream in junhoyeo/tokscale#805, that's the part that actually matters here.

Two notes on the Token Monitor side:

- `~/.omp/agent/sessions` is already scanned natively by current tokscale, so the `TOKSCALE_EXTRA_DIRS=pi:...` line isn't needed.
- The real blocker for #46 is exactly that title-line case: `pi.rs` hits the leading `title` line, `entry_type != "session"` is true, and `return Vec::new()` drops the whole file. That parser fix isn't in a released tokscale yet though. 4.0.8 predates it, and 4.0.9 / 4.0.10 still have the `return Vec::new()`, so the 4.0.8 bump here doesn't resolve OMP counting on its own.

I'd rather not merge as-is, but I'll keep this open and track it upstream. Once the fix lands in a tokscale release, we can update this PR to bump to that version so the change actually carries the fix. Thanks again for the upstream work.

PR #55, opened 2026-07-03
fix(tray): respect limit bar modes in menu bar text and icons

atukunare:
## Summary
- Show the worst remaining percent in menu bar text for `bars`, `barsSession`, `barsWeekly`, and `barsAllSessions` instead of hiding text when limit data exists.
- Route tray bar icon rendering through shared `trayText` helpers so session/weekly/billing windows fall back correctly (helps Cursor/Antigravity and billing-only providers).
- Add regression tests for tray limit text and fallback behavior.

## Test plan
- [x] `node --test tests/electron/tray.test.js`
- [ ] macOS: set tray content to Session bars / Weekly bars and confirm text + icon update for Cursor/Antigravity accounts
- [ ] macOS: confirm token/cost modes unchanged

### DIFF
--- a/src/electron/renderer/app.js
+++ b/src/electron/renderer/app.js
@@ -5222,19 +5222,6 @@ window.tokenMonitor.onStatsPush?.((payload) => {
   restartTimer();
 });
 
-function pickWorstProvider(stats, windowFilter) {
-  const pair = window.TokenMonitorTrayText.pickWorstBarPair(stats, windowFilter || null);
-  return pair?.provider || null;
-}
-
-function pickWorstSessionProvider(stats) {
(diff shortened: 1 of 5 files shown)


Javis603:
Thanks @atukunare. The main change here, showing the worst remaining percent as text in the `bars` / `barsSession` / `barsWeekly` / `barsAllSessions` modes, is something I'd rather not take: those modes are intentionally icon-only, the bars already carry the number, so adding text changes the menu bar footprint for everyone already on them. If we want a number-in-text option, I'd lean toward a distinct mode rather than changing what the existing modes do.

It's also drifted against main since it was opened, with a fair number of new providers and tray changes landing in between. I'm going to close this one, but thanks for taking a run at it.

PR #56, opened 2026-07-03
feat(macos): add launchd helpers for dev auto-start

atukunare:
## Summary
- Add `scripts/launchd-{install,uninstall,status}.sh` and `scripts/launch-token-monitor.sh` for macOS dev workflows.
- Expose npm scripts: `launchd:install`, `launchd:uninstall`, `launchd:status`.
- Document optional login auto-start in README (with `TOKEN_MONITOR_APP` override for packaged `.app` builds).

## Test plan
- [ ] macOS: `npm run launchd:install` → log in again → widget starts hidden from source checkout
- [ ] `npm run launchd:status` shows loaded job
- [ ] `npm run launchd:uninstall` removes LaunchAgent

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,16 @@ npm install
 npm start
 ```
 
+**macOS dev auto-start (optional):** register a login LaunchAgent so a source checkout starts hidden on sign-in:
+
+```bash
+npm run launchd:install   # install LaunchAgent
+npm run launchd:status    # check status
+npm run launchd:uninstall # remove
+```
+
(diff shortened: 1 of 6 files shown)


Javis603:
Thanks @atukunare for putting this together.

Coming to this after a while: for end users the packaged app already has login auto-start built in (the Start at login setting, via `setLoginItemSettings`), so this really covers the narrower case of running from a source checkout. That's a pretty niche dev convenience to carry as four shell scripts plus npm entries and a README section, and macOS-only, so I'd rather not add it to the repo.

Closing this one, but thanks for the clean, self-contained implementation.

PR #57, opened 2026-07-03
feat(i18n): add Korean locale to the desktop app

atukunare:
## Summary
- Add **한국어** to Settings → Language.
- Port 466 Korean strings from community work; keys added upstream since v0.15 (Copilot/Kiro/Home modules/etc.) fall back to English until translated.
- Include `scripts/contrib/merge-locale-into-upstream-i18n.js` to help refresh the locale against future English key additions.

## Test plan
- [ ] Settings → Language → Korean updates the UI immediately
- [ ] Auto (system) picks `ko-KR` on Korean macOS/Windows
- [ ] Spot-check Limits, Sync, and Home views for untranslated fallback strings


Javis603:
Thanks for adding the Korean locale, this is a useful addition!

Two things before I merge. A few strings render the product name as "Token Mate", but this project ships as "Token Monitor". Could you update these to match?

- `floatingBubble.expand`
- `settings.startup.launchAtSignIn`
- `settings.sync.hostWarning`
- `settings.tokscale.noteBefore`

Also I'd rather not carry `scripts/contrib/merge-locale-into-upstream-i18n.js` in the repo, it is a one-off maintenance helper that most contributors won't need. Would you mind dropping it from the PR?

Once those are in I think this is in good shape.

atukunare:
> Thanks for adding the Korean locale, this is a useful addition!
> 
> Two things before I merge. A few strings render the product name as "Token Mate", but this project ships as "Token Monitor". Could you update these to match?
> 
> * `floatingBubble.expand`
> * `settings.startup.launchAtSignIn`
> * `settings.sync.hostWarning`
> * `settings.tokscale.noteBefore`
> 
> Also I'd rather not carry `scripts/contrib/merge-locale-into-upstream-i18n.js` in the repo, it is a one-off maintenance helper that most contributors won't need. Would you mind dropping it from the PR?
> 
> Once those are in I think this is in good shape.

Thanks for the review — both points are addressed in the latest push.
- Updated the four Korean strings to say "Token Monitor" instead of "Token Mate".
- Removed `scripts/contrib/merge-locale-into-upstream-i18n.js` from the PR.
Let me know if anything else needs tweaking.

Javis603:
Thanks for turning both points around so quickly.

I pushed one commit (1fcff3e) before merging that finishes wiring up the locale: the dropdown option (index.html) and the main-process language allowlist (main.js) both still needed `ko`, so I added the option and derived the allowlist from LANGUAGE_OPTIONS so future locales stay in sync. It also translates the strings that were still falling back to English and lists the option as 한국어 in the Chinese menus, with a couple of tests so a locale can't be half-wired again.

PR #58, opened 2026-07-03
feat(i18n): add Japanese locale to the desktop app

atukunare:
## Summary
- Add **日本語** to Settings → Language.
- Port 466 Japanese strings from community work; newer upstream-only keys fall back to English.
- Reuses `scripts/contrib/merge-locale-into-upstream-i18n.js` for future locale refreshes.

## Test plan
- [ ] Settings → Language → Japanese updates the UI immediately
- [ ] Auto (system) picks `ja-JP` on Japanese macOS/Windows

Made with [Cursor](https://cursor.com)


Javis603:
Thanks for adding the Japanese locale, this is a useful addition!

Two things before I merge. A few strings render the product name as "Token Mate", but this project ships as "Token Monitor". Could you update these to match?

- `floatingBubble.expand`
- `settings.startup.launchAtSignIn`
- `settings.sync.hostWarning`
- `settings.tokscale.noteBefore`

Also I'd rather not carry `scripts/contrib/merge-locale-into-upstream-i18n.js` in the repo, it is a one-off maintenance helper that most contributors won't need. Would you mind dropping it from the PR?

Once those are in I think this is in good shape.

atukunare:
> Thanks for adding the Japanese locale, this is a useful addition!
> 
> Two things before I merge. A few strings render the product name as "Token Mate", but this project ships as "Token Monitor". Could you update these to match?
> 
> * `floatingBubble.expand`
> * `settings.startup.launchAtSignIn`
> * `settings.sync.hostWarning`
> * `settings.tokscale.noteBefore`
> 
> Also I'd rather not carry `scripts/contrib/merge-locale-into-upstream-i18n.js` in the repo, it is a one-off maintenance helper that most contributors won't need. Would you mind dropping it from the PR?
> 
> Once those are in I think this is in good shape.

Thanks for the review — both points are addressed in the latest push.
- Updated the four Japanese strings to say "Token Monitor" instead of "Token Mate".
- Removed `scripts/contrib/merge-locale-into-upstream-i18n.js` from the PR.
Let me know if anything else needs tweaking.

Javis603:
Thanks for turning both points around so quickly.

I pushed one commit (013806a) before merging that finishes wiring up the locale: the dropdown option (index.html) and the localeFromTag resolver both still needed `ja` (the main-process allowlist already derives from LANGUAGE_OPTIONS so it picked it up), so I added those and list the option as 日本語 in the Chinese menus. It also translates the strings that were still falling back to English, and the tests added for the Korean locale now cover Japanese too.

PR #63, opened 2026-07-04
feat: add native ZCode usage tracking via SQLite

Ankairis:
## Summary

ZCode stores per-request token usage in `~/.zcode/cli/db/db.sqlite` (`model_usage` table), **not** in `~/.zcode/projects/` (which rarely exists on real installs). The current ZCode support is effectively broken — tokscale scans a non-existent directory, so ZCode always shows as waiting/missing with 0 tokens.

This PR adds a direct `node:sqlite` reader that bypasses tokscale entirely for ZCode, following the same pattern already used for OpenCode (`opencodeSession.js`).

### Core: `src/shared/zcodeUsage.js` (new)
- Read-only `node:sqlite` reader of `~/.zcode/cli/db/db.sqlite::model_usage`
- Terminal-state filter: only sync rows where `completed_at IS NOT NULL AND status IN ('completed','cancelled','error')` and at least one token > 0 (prevents half-written streaming rows from being permanently undercounted)
- Maps each row to a neutral format consumable by `extractUsageFromTokscale`
- Bucketing into today/month/allTime by `completed_at` against local midnight boundaries
- Cache-inclusive semantics preserved: `input_tokens` already contains `cache_read`; `reasoning` kept as informational metadata (not added to output — output is already pure)
- Graceful degradation when `node:sqlite` unavailable or db missing/unreadable (returns null, caller treats as empty)

### Integration: `src/shared/collector.js`
- After tokscale scans, if `zcode` is in enabled clients, call `collectZcodeUsage()` and `mergePeriods` the result into today/month/allTime
- Watch path corrected from `~/.zcode/projects` to `~/.zcode/cli/db` (chokidar sees db + -wal/-shm sidecar changes for seconds-level refresh on completed requests)
- Added explanatory comments referencing cc-switch's `zcode-usage-tracking.md`

### WSL: `src/shared/wslUsage.js`
- `WSL_DATA_MARKERS`: `.zcode/projects` changed to `.zcode/cli/db`
- `MARKER_CLIENTS` key updated accordingly

### Docs: README (EN/zh-CN/zh-TW)
- ZCode data path corrected: `~/.zcode/projects/` changed to `~/.zcode/cli/db/db.sqlite (model_usage)`

### Tests
- `tests/shared/zcodeUsage.test.js`: 7 tests covering terminal-state filter, cache read/write split, reasoning exclusion, today/month/allTime bucketing, missing db handling, session metadata, provider mapping — **all pass**
- `tests/shared/wslUsage.test.js`: updated marker path assertion — **all 22 tests pass**

## Known Limitations

- **WSL direct read**: ZCode usage inside WSL still goes through tokscale (which scans the non-existent `~/.zcode/projects` and returns 0). Direct SQLite read across `\\wsl$` paths is out of scope for this usage-only PR. Documented in collector.js comment.
- **Cost**: ZCode rows carry cost=0 (token-monitor's pricing module is tokscale-based; ZCode pricing is not in scope). Rows still write correctly for token statistics.
- **Session Details / AI Tool Limits**: Out of scope (this PR is usage tracking only, matching the cc-switch reference doc scope).

## Verification

```bash
node --test tests/shared/zcodeUsage.test.js   # 7/7 pass
node --test tests/shared/wslUsage.test.js      # 22/22 pass
```

Tested against real ZCode data: `~/.zcode/cli/db/db.sqlite` on local machine contains 1038 completed + 36 cancelled + 30 error rows — all picked up correctly.

## Design Decisions

| Decision | Rationale |
|---|---|
| Direct SQLite read, not tokscale | tokscale scans `~/.zcode/projects` (doesn't exist); OpenCode precedent proves this approach is accepted in token-monitor |
| Terminal-state only SQL filter | Per cc-switch: streaming rows have half-written tokens; early sync leads to permanent undercount via INSERT OR IGNORE |
| Cost = 0 | token-monitor's pricing is tokscale-based; ZCode pricing out of scope |
| No cache_read subtraction from input | token-monitor's total adds cache_read as independent component (consistent with OpenCode/Zed/MiMo) |
| Watch `~/.zcode/cli/db` dir | Real write location; ensures seconds-level refresh and correct active status |


Ankairis:
Closing this PR — on investigation, tokscale 4.x (already a dependency of token-monitor) natively reads ZCode's SQLite db at `~/.zcode/cli/db/db.sqlite` via its `additionalPaths` mechanism (`tokscale clients --json` shows `client: zcode`, `messageCount: 1182`). The repo's existing ZCode support was therefore already functional end-to-end.

My direct `node:sqlite` reader in this PR was redundant and double-counted ZCode usage on top of tokscale's own scan (`mergePeriods` summed both sources). The \"watch path\" theory and the \"tokscale scans ~/.zcode/projects which doesn't exist\" premise were both wrong — tokscale uses the SQLite db directly regardless of the legacy sessionsPath.

No code changes needed. Apologies for the noise, and thanks for the reviews.

PR #65, opened 2026-07-04
ci: add Linux AppImage build

freefrank:
## Summary
Adds Linux **AppImage** support so Token Monitor has a runnable Linux build alongside macOS/Windows. This is the smallest change that produces a Linux artifact without touching the mac/win pipeline.

## What changed
- **`package.json`** — a `build.linux` electron-builder target (`AppImage`, x64) plus `dist:linux` / `predist:linux` scripts, mirroring the existing `dist:mac` / `dist:win`. Icon reuses the generated `build/icons` set.
- **`.github/workflows/appimage.yml`** — a dedicated workflow that builds on push/PR (build-health check) and, on `v*` tags, attaches the `.AppImage` to the GitHub release.

## Notes
- `APPIMAGE_EXTRACT_AND_RUN=1` is set on the build step so electron-builder's bundled `appimagetool` runs without `libfuse2` (not present on the runners, and increasingly absent on modern desktops too).
- **x64 only** for now — arm64 can follow if there's demand.
- Kept as a **standalone workflow** to avoid touching `release.yml`. If you'd rather fold Linux into the existing `release.yml` matrix (one release job for mac/win/linux), I'm happy to rework it that way — just say which you prefer.
- **Out of scope here:** "Start at login" stays disabled on Linux (unchanged from `main`). Electron's `setLoginItemSettings` is macOS/Windows-only, so a Linux autostart would need a `~/.config/autostart` fallback — opened #66 to confirm whether that's wanted before sending code.

## Testing
- Built and ran locally on Fedora (GNOME/Wayland): produces `dist/Token Monitor-0.20.0.AppImage` (~145 MB); the app launches and the collector picks up `~/.claude/projects` and `~/.codex/sessions`.
- Green in CI on my fork (~45s, artifact uploaded): https://github.com/freefrank/token-monitor/actions/runs/28715346980


Javis603:
Thanks, clean split. Your Fedora run and the green fork CI cover the Linux side I hadn't verified myself.

I pushed one commit (ebfb8d3) before merging that folds the build into release.yml's existing mac/win matrix and drops the standalone appimage.yml: with both workflows attaching to the same `v*` release they'd race to create it and could clobber the release body, so one release job now publishes all three platforms. Your `build.linux` target and the `APPIMAGE_EXTRACT_AND_RUN` handling carried over unchanged; I also bumped that workflow to Node 22 (it was still on 20) to match engines and the version you built on, and kept it x64-only as you had it.

PR #83, opened 2026-07-07
feat: add hover details to home activity heatmap

### LINKED ISSUE #75: feat希望能添加活动区域内格子的悬浮显示
Image

Image

希望能像codex等工具一样显示当日token消耗等详情

joyehuang:
## Summary
- add per-day token/cost data to Home activity heatmap cells
- show an immediate custom tooltip with date and token usage on hover
- add a smooth spotlight/highlight effect for hovered activity cells

Closes #75

## Tests
- npm run lint
- npm test -- tests/electron/usageCharts.test.js tests/electron/homeOverview.test.js

### DIFF
--- a/src/electron/renderer/app.js
+++ b/src/electron/renderer/app.js
@@ -2626,6 +2626,20 @@ function setupHomeActivityHover(scroller) {
   });
   scroller.addEventListener('pointerleave', hide);
   scroller.addEventListener('scroll', hide);
+  // The tooltip lives on document.body and is only dismissed by handlers on this
+  // scroller, which renderHome() throws away on every rebuild. Expose the latest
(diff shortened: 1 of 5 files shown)


Javis603:
Thanks @joyehuang, the tooltip and spotlight both look great!

I pushed one follow-up commit (a5a8226) before merging that fixes the tooltip lifecycle: the tooltip is attached to `document.body`, and only the scroller's own pointer handlers dismissed it. Since `renderHome()` rebuilds that scroller on every stats refresh, and leaving Home only hides the panel, a cell hovered across a refresh or view switch could leave a stale tooltip on screen. The fix exposes the hover teardown and calls it from `renderHome()` and whenever the active view is not Home.

PR #111, opened 2026-07-10
fix(limits): prefer fresh synced provider states

joyehuang:
## Summary

- prefer fresh provider observations over stale records from old or renamed devices
- keep stale quota data visible only when no fresh observation for that provider exists
- apply the same aggregation behavior in the Node hub and generated Worker copy

## Root cause

Configured accounts and unconfigured provider states can have different aggregation keys. An orphaned stale device could therefore keep a configured quota candidate while the current device reported a fresh `notConfigured` state, and the later provider collapse could select the stale quota instead of the current observation.

The aggregation pass now records which providers have fresh observations and excludes stale candidates whenever a current observation exists. If every observation is stale, the most useful stale quota remains visible as before.

## Impact

Synced dashboards reflect the current device/provider state instead of showing quota data left behind by a stale device id.

## Validation

- `npm run verify` (lint + 1028 tests)

---
## Summary by cubic
Prefer fresh provider states during limits aggregation so stale data from old or renamed devices can’t override the current state. Dashboards now reflect the current device/provider status.

- Bug Fixes
  - Ignore stale candidates when a fresh observation for that provider exists.
  - If fresh data exists, only consider configured accounts from fresh devices; otherwise keep the most useful stale configured quota.
  - Apply the same logic in `src/shared/limits.js` and `worker/src/shared/limits.js`, with tests covering these cases.

<sup>Written for commit 05a60acc764c60985ec0012134bb9283948f3abe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/111?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/shared/limits.js
+++ b/src/shared/limits.js
@@ -339,16 +339,22 @@ function aggregateLimits(devices, staleAfterMs = 0, nowMs = Date.now()) {
   const aggregate = { updatedAt: new Date(nowMs).toISOString(), providers: [] };
   const byKey = new Map();
   const providersWithConfiguredAccounts = new Set();
+  const providersWithFreshConfiguredAccounts = new Set();
+  const providersWithFreshObservations = new Set();
 
   for (const device of devices || []) {
     const summary = normalizeLimitsSummary(device?.limits);
     for (const provider of summary.providers) {
-      if (isConfiguredProvider(provider)) providersWithConfiguredAccounts.add(provider.provider);
       const candidate = {
         ...provider,
         sourceDeviceId: String(device?.deviceId || ''),
         stale: isProviderStale(provider, summary, device, staleAfterMs, nowMs)
       };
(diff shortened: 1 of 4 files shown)


Javis603:
Thanks @joyehuang for the fresh synced provider state fix!

The new precedence is right: a fresh `notConfigured` observation should win over an orphaned stale quota, while stale provider data remains available when it is the only observation. The code and tests look good.

PR #112, opened 2026-07-10
fix(limits): present reset windows accurately

joyehuang:
## Summary

- stop rendering an expired quota timestamp as `Reset now` indefinitely
- show `No active 5h window` when Claude reports an unused session window without a reset timestamp
- keep the Limits and Home quota presentations consistent, with translations for every bundled locale

## Root cause

The renderer treated every timestamp in the past as `Reset now`, regardless of how old it was. A stale provider timestamp could therefore remain visible forever. Claude can also legitimately return a zero-utilization five-hour window without `resets_at`; that state was rendered as a blank reset line instead of an inactive window.

The new presentation helper keeps a one-minute grace period around the reset boundary, then drops expired timestamps. It also distinguishes Claude's inactive five-hour window from a used window whose reset timestamp is unexpectedly missing.

## Impact

Quota cards no longer show misleading permanent reset labels, and an inactive Claude five-hour window is explained instead of appearing broken.

## Validation

- `npm run verify` (lint + 1030 tests)

---
## Summary by cubic
Fixes inaccurate quota reset labels and inactive session handling. Reset windows now expire cleanly, reset durations are localized, and unused Claude 5h sessions show a clear message across Limits and Home.

- **Bug Fixes**
  - Added `limitResetRemainingMs` with a 1‑minute grace; localized "Reset now" and duration strings; dropped stale timestamps.
  - Unified reset rendering in Limits and Home; only use fallback text when `resetsAt` is missing.
  - Introduced `isInactiveLimitWindow` for Claude sessions; unused 5h windows show "No active 5h window."
  - Added i18n keys for days/hours/minutes/<1m in all locales and tests for duration formatting, expiry, inactive vs. used, and shared rendering.

<sup>Written for commit 24eee79f26e39f771341bf4826204464ab893f61. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Javis603:
Thanks @joyehuang for identifying the stale reset-window behavior and for the follow-up work. I carried the fix into #212 and retained your authorship on the relevant commit. I rebuilt it on current main with source-owned, provider/account-scoped reset-boundary refresh, while keeping the existing reset copy, duration formatting, and Claude idle-window semantics.

I'm closing this PR as superseded by #212.

PR #116, opened 2026-07-11
fix(limits): stabilize Codex quota refreshes

Javis603:
## Summary

- prefer the canonical Codex quota snapshot and refuse to guess between conflicting alternate rate-limit buckets
- retain recent same-account quota windows only across empty or transient reads, including the identity-less single live-account failure path
- trust successful non-empty snapshots so upstream global resets, reset credits, and reset-time changes appear immediately
- keep account-scoped refreshes merged with the latest limits snapshot and sync the shared behavior to the Worker copy

## Testing

- `node --test tests/shared/limits.test.js tests/shared/limitCollector.codex.test.js`
- `npm run verify` (1069 tests)

---
## Summary by cubic
Stabilizes Codex quota refreshes by trusting canonical snapshots, validating fallback buckets, and retaining recent windows across transient reads. Prevents flicker and keeps managed-account refreshes consistent across the app and the Worker.

- **Bug Fixes**
  - Validate fallback snapshots: only use alternate limit IDs when their windows match; ignore ordering; keep `planType` and reset-credits only if every alternate agrees; never choose between conflicting alternates.
  - Retain recent Codex windows for up to 10 minutes when the current read is empty or transient (`unavailable`, `error`, `rateLimited`, `sourceRateLimited`), including the single live-account path with no identity; never guess identity when ambiguous.
  - Treat successful non-empty snapshots as authoritative so quota increases and reset-target changes take effect immediately; do not backfill older windows into them.
  - Apply `mergeCodexTransientWindows` in managed account refreshes and mirror the same logic in the Worker to keep behavior consistent (including stable timestamps during retention).

<sup>Written for commit bc62ae2479b35063acf608bae61e0defc4900c54. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/116?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -21,6 +21,7 @@ const { customPricingPath } = require('../shared/tokscaleConfig');
 const { applyCustomPricing, normalizeCustomPricingSetting } = require('../shared/tokscaleCustomPricing');
 const { createHub } = require('../hub/server');
 const { collectLimitsOnce, deepseekToken, normalizeLimitsRefreshMs, parseBoolean, parseLimitProviders, runCodexLogin, minimaxToken, copilotToken, zaiToken, zaiRegion, zaiTeamToken, volcengineCredentials, qoderCookie, kimiToken } = require('../shared/limitCollector');
+const { mergeCodexTransientWindows } = require('../shared/limits');
 const { copilotLoginErrorMessage, isAllowedVerificationUrl, runCopilotDeviceFlowLogin } = require('../shared/copilotDeviceFlow');
 const { codexAuthIdentity, hashAccountKey } = require('../shared/codexAuth');
 const { codexLoginUrlFromOutput, isAllowedCodexLoginUrl } = require('../shared/codexLogin');
@@ -855,9 +856,10 @@ async function refreshCodexManagedAccountLimits(id) {
       includeLiveCodexAccount: false,
       codexManagedAccounts: [account]
     }, { env: process.env });
+    const stableSummary = mergeCodexTransientWindows(latestStats?.limits, summary);
     return {
       ok: true,
-      providers: (summary.providers || []).filter((provider) => provider?.provider === 'codex')
+      providers: (stableSummary.providers || []).filter((provider) => provider?.provider === 'codex')
     };
   } catch (error) {
     return { ok: false, error: `Could not refresh Codex account limits: ${error?.message || error}` };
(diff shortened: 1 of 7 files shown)


joyehuang:
I can still reproduce a separate cross-device flicker after the single-device refresh path is stabilized here.

Two fresh devices report the same Codex `accountKey`, but the app RPC returns different active quota buckets:

| Device | Session used | Weekly used | Reset anchors |
| --- | ---: | ---: | --- |
| Windows | 4% | 1% | later session/weekly anchors |
| macOS | 88% | 14% | earlier session/weekly anchors |

Both snapshots are `status: ok`, `source: rpc`, and `sourceDetail: app`. `aggregateLimits()` currently collapses them by account and selects whichever provider has the newest `updatedAt`, so alternating device reports make the UI jump between the two valid snapshots. This is downstream of the collector behavior addressed by this PR; neither read is empty or transient.

I tested a focused aggregation-layer fix locally:

- apply only when two Codex providers have the same non-empty `accountKey`;
- preserve the existing stale/status/window ranking first;
- among active, non-expired quota snapshots, deterministically keep the most constrained real snapshot rather than the most recently reported device;
- drop expired windows from that comparison so a genuine new quota cycle is accepted immediately;
- leave all other providers on the existing freshest-snapshot behavior.

The regression coverage checks reversed device order, equal percentages with different reset anchors, and transition after the tighter window expires. It also keeps the selected snapshot intact (rather than synthesizing per-window values), so `sourceDeviceId` remains truthful.

Would you prefer this folded into #116, or should I send it as a small follow-up PR rebased onto `upstream/main` after #116 lands?

Javis603:
Thanks @joyehuang for isolating this. A separate follow-up PR rebased onto `main` after #116 lands would be better. This is a different policy boundary: choosing between two successful cross-device snapshots rather than retaining a failed or empty read, so keeping it separate will make the aggregation rule and regression coverage easier to review.

The same-account scope and whole-snapshot selection make sense. Please also cover an upstream/global reset or a user-triggered reset-credit action arriving on one device before the older, tighter snapshot's `resetAt`; aggregation should accept the new quota generation immediately rather than hold the old snapshot until expiry. For the reset-credit case, please do not treat an `availableCount` decrease by itself as proof of a manual reset, since an unused credit can also expire. Require the quota windows to move to a new reset cycle too.

PR #125, opened 2026-07-11
fix(limits): stabilize cross-device Codex quota

joyehuang:
## Summary

- stabilize same-account Codex limits when multiple devices report different successful quota snapshots
- prefer the tightest whole snapshot within one reset generation instead of whichever device reported last
- accept a newer quota generation immediately when comparable reset anchors advance, including upstream resets and reset-credit actions
- keep reset-credit counts out of generation detection so an expiring unused credit cannot mimic a manual reset

## Why

Two fresh devices can report the same Codex `accountKey` while returning different active quota buckets. The hub currently collapses those rows by account and selects the newest `updatedAt`, so alternating device reports can make the UI jump between substantially different session and weekly percentages.

This is separate from the transient or empty-read handling in #116: both cross-device snapshots are successful and non-empty. The aggregation rule now remains deterministic without retaining historical state or synthesizing windows from different devices, so `sourceDeviceId` continues to identify the selected whole snapshot.

## Reset handling

Comparable active reset anchors define the quota generation. A generation wins only when its anchors move in one direction with at least one advancing and none moving backward. This lets a new upstream/global reset or a reset-credit action replace an older tighter snapshot before its previous reset time.

`resetCredits.availableCount` is not used as reset evidence. The count may decrease when an unused credit expires, so a quota-cycle movement is still required.

## Verification

- `npm run verify` (1119 tests passed)
- `node --test tests/shared/limits.test.js tests/shared/usage.test.js` (65 tests passed)
- `git diff --check`

The regression coverage includes reversed device order, same-generation conflicts, an upstream reset before expiry, reset-credit expiry without a cycle change, a reset-credit action with a cycle change, and normal transition after expiry.

---
## Summary by cubic
Stabilizes cross‑device Codex quota aggregation and makes it order‑independent. We keep the tightest active snapshot per account and switch only when a newer reset generation is detected; equal‑pressure snapshots now resolve consistently.

- **Bug Fixes**
  - Deterministic same‑`accountKey` selection: build a generation frontier across device snapshots, then pick the tightest active whole snapshot; device report order no longer changes the result.
  - Generation changes: when comparable reset anchors advance, accept the newer generation immediately (upstream resets or reset‑credit actions); ignore `resetCredits.availableCount` alone and skip jointly expired anchors.
  - Active windows only: rank and compare using only active limit windows.
  - Stable ties: for equal‑pressure same‑account Codex snapshots, break by `sourceDeviceId`, then `updatedAt`, then a final structural tiebreaker; updated `docs/API.md` and added tests for order permutations, equal‑pressure refreshes, mixed‑window frontier, expired session with active weekly, jointly expired anchors, credit expiry without a cycle move, and normal cycle transitions.

<sup>Written for commit adf0087ccf271c4e949b9119f6bd16e5a5d1ca0e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/125?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -198,7 +198,7 @@ Response includes:
 - `devices`
 - stale status for devices that have not reported recently
 
-If multiple devices report the same provider account, the hub normally keeps the freshest valid limits status for that account. Codex is the exception: simultaneous app sessions can report different active quota buckets for the same account, so the hub keeps the tightest whole snapshot within a reset generation. When comparable reset anchors advance together, the newer generation wins immediately, including after an upstream reset or reset-credit action. Public Worker stats omit account identifiers.
+If multiple devices report the same provider account, the hub normally keeps the freshest valid limits status for that account. Codex is the exception: simultaneous app sessions can report different active quota buckets for the same account, so the hub keeps the tightest whole snapshot within a reset generation. When at least one comparable reset anchor advances and none regress, the newer generation wins immediately, including after an upstream reset or reset-credit action. Public Worker stats omit account identifiers.
 
 ## `GET /api/devices`
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -198,7 +198,7 @@ Response includes:
 - `devices`
 - stale status for devices that have not reported recently
 
-If multiple devices report the same provider account, the hub keeps the freshest valid limits status for that account. Public Worker stats omit account identifiers.
+If multiple devices report the same provider account, the hub normally keeps the freshest valid limits status for that account. Codex is the exception: simultaneous app sessions can report different active quota buckets for the same account, so the hub keeps the tightest whole snapshot within a reset generation. When comparable reset anchors advance together, the newer generation wins immediately, including after an upstream reset or reset-credit action. Public Worker stats omit account identifiers.
 
 ## `GET /api/devices`
 
--- a/src/shared/limits.js
+++ b/src/shared/limits.js
@@ -345,11 +345,10 @@ function providerWindowRank(provider, nowMs) {
   return Array.isArray(provider.windows) && provider.windows.some((window) => activeLimitWindow(window, nowMs)) ? 1 : 0;
 }
 
-function codexResetGeneration(provider, nowMs) {
+function codexResetGeneration(provider) {
   if (provider?.provider !== 'codex' || !Array.isArray(provider.windows)) return null;
   const generation = new Map();
   for (const window of provider.windows) {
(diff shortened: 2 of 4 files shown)


Javis603:
Thanks @joyehuang for tackling the cross-device Codex quota oscillation. I agree with the direction: inconsistent readings originate upstream, but this remains a useful defensive layer because valid snapshots from multiple devices can arrive delayed or out of order.

I found one correctness issue before merge. `codexResetGeneration()` currently includes every reset anchor. For example, at 13:00, if both devices have expired session anchors pointing in the opposite direction to an active weekly anchor that advanced from 7/18 to 7/19, `compareCodexResetGeneration()` treats the result as ambiguous. Quota pressure then re-selects the older, tighter weekly snapshot, suppressing a real weekly or global reset.

The fix needs to be narrower than dropping inactive windows per snapshot: when comparing the same anchor, skip it only if it is expired in both snapshots. In other words, thread `nowMs` through the comparison and `continue` when `firstReset <= nowMs && secondReset <= nowMs`. Dropping inactive windows independently breaks the existing `advances an expired Codex session cycle while the weekly window remains active` test because that case needs the old expired anchor to recognize the new cycle.

Please add a regression test for the both-expired case. Everything else looks solid, and the focused suite plus `npm run verify` pass on the current head.

Javis603:
Thanks @joyehuang for the thorough work here, especially the follow-up for the jointly expired reset-anchor case.

During the first few days after GPT-5.6 rolled out, I did personally observe the same Codex account reporting different quota values across devices and jumping between them. That was why I initially thought a defensive aggregation layer was worth keeping and asked for the follow-up fix.

Since then, the upstream behavior appears to have stabilized, and I haven't seen the issue recur for several days, close to a week. After revisiting the maintenance trade-off, I don't think Token Monitor should permanently encode a “tightest snapshot wins” policy for this. It could stabilize an incorrect low reading just as easily as a correct one, while adding substantial reset-generation and ordering logic to the shared hub path.

I'm going to close this one rather than carry that complexity without a current reproducible case. Thank you again for taking the implementation and regression coverage this far.

PR #148, opened 2026-07-14
feat(sync): add configurable upload frequency

leoz9:
## Summary

- add a sync upload frequency setting with Live, 10, 20, and 30 minute choices
- coalesce high-frequency summaries while serializing remote-hub uploads
- keep local usage and AI Tool Limits live in client mode without waiting for the next Hub upload
- expose the Hub's effective `staleAfterMs` and retain device `periodWindows` so client-side aggregation preserves provider freshness and calendar boundaries
- extend device and limit staleness thresholds for interval-synced devices, with a conservative fallback for older Hubs
- document and test the sync wire fields

## Why

The desktop collector can produce summary updates multiple times per second. Multi-device sync previously forwarded every update to the configured Hub, which could create unnecessary Cloudflare Worker traffic for users on the free tier.

Live remains the default and preserves the previous upload behavior. A fixed interval uploads the first summary immediately, then coalesces later updates so only the newest pending summary is sent when the interval expires. Local rendering remains driven by the collector, so reducing remote upload frequency does not delay this device's usage or Limits. Failed uploads do not throttle a newer pending summary.

## Validation

- `npm run verify` (1,351 passed)
- targeted sync, usage, Limits, Hub, and Worker tests (96 passed)
- `npm run sync:worker`
- `git diff --exit-code worker/src/shared`
- `git diff --check`


Javis603:
Thanks @leoz9 for adding the configurable sync upload cadence. The Live default preserves the existing behavior, while fixed intervals coalesce collector updates and reduce unnecessary Hub traffic.

I pushed maintainer follow-ups through 1b7e9f44 to keep local rendering independent from the remote upload cadence: local usage and AI Tool Limits update directly from the collector, Hub freshness and period-window metadata are preserved during client-side aggregation, and local updates no longer mask an offline Hub connection. The corresponding env default is now documented in `.env.example`.

PR #158, opened 2026-07-15
fix(updater): restore dismissed update actions

Javis603:
## Summary

- keep update availability independent from whether the footer notification was dismissed
- restore the same-version notification after a manual update check or user-requested download
- keep the Settings download action available after dismissing the footer pill
- turn a downloaded update into a direct **Restart to update** action instead of reopening release notes
- hide the dismiss control while an update is downloading or ready to install, and localize its accessibility labels

## Behavior

| State | Footer pill | Settings action |
| --- | --- | --- |
| Update available | Shows release notes before download | Download update |
| Notification dismissed | Hidden until a manual check or newer release | Download update remains available |
| Downloading | Shows progress without a dismiss action | Disabled while busy |
| Downloaded | Direct Restart to update action | Restart to update |

The updater state is shared by macOS, Windows, and supported Linux AppImage builds, so the fix applies consistently across platforms.

## Testing

- `npm run verify` (1304 tests)

---
## Summary by cubic
Fixes the app update flow so dismissing the footer pill doesn’t block updates. Manual checks or starting a download restore access, and downloaded updates become a direct “Restart to update” action.

- **Bug Fixes**
  - Kept update availability separate from dismissal via `deriveAppUpdateAvailability` (adds `showUpdateNotice` and `downloaded`) and reused it for cooldown selection.
  - Manual checks await any in-flight background check, publish the checking state immediately, then restore a matching dismissed version and surface failures if the shared request fails.
  - Starting a user-requested download also restores a dismissed version.
  - Downloaded updates install directly (↻ Restart) without reopening release notes; the pill bypasses the popover in install mode and hides the dismiss control while downloading or ready.
  - Kept the Settings “Download update” action available even when the footer pill is dismissed; localized dismiss button labels.
  - Added tests for availability, dismissal restore, joined-check feedback, in-flight check reuse, and pill behavior; applies on macOS, Windows, and Linux AppImage.

<sup>Written for commit c8eee470febbd947bb20d9ba9165bf7b38274ba6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/158?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -2877,6 +2877,7 @@ async function downloadTokscaleFromNpm() {
 }
 
 let appUpdateCheckInFlight = false;
+let appUpdateCheckPromise = null;
 let appUpdateLastError = null;
 let appUpdateBackgroundTimer = null;
 let appUpdateNativeBusy = false;
@@ -3004,7 +3005,14 @@ function sendAppUpdatePush() {
 }
 
 async function runAppUpdateCheck({ force = false } = {}) {
-  if (appUpdateCheckInFlight) return deriveAppUpdateState();
+  if (appUpdateCheckPromise) {
+    const activeResult = await appUpdateCheckPromise;
+    if (force && activeResult?.ok && activeResult.newer) {
+      restoreDismissedAppUpdate(activeResult.latest?.version);
+      sendAppUpdatePush();
+    }
+    return deriveAppUpdateState();
+  }
   const block = settings?.appUpdate || {};
   if (shouldSkipAppUpdateCheck({
     force,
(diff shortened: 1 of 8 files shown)


Javis603:
Addressed all three findings in be891d3c: the cooldown now uses the shared availability predicate, manual checks await an in-flight background check before restoring the dismissed release, and the source-scanning tests now guard their slice boundaries. `npm run verify` passes with 1305 tests.

PR #160, opened 2026-07-15
feat(collector): add activity-gated smart collection

wetlink:
## Summary

- add a `Smart (10 minutes)` collection cadence
- use native filesystem events to record agent activity without collecting immediately
- coalesce activity until the next cadence tick instead of scanning for every file event
- retain an unconditional hourly full reconciliation as a recovery path
- keep the existing live and fixed-interval modes backward compatible

This addresses the idle CPU and disk pressure discussed in #52 while keeping filesystem events as an optimization rather than the sole source of truth.

## Smart mode contract

- performs the normal initial collection on startup
- watches supported host-side agent directories with native filesystem events
- runs an anchored collection at the next 10-minute tick when activity is pending
- skips clean 10-minute ticks when no activity is pending
- always performs the existing hourly full scan, even when no event was observed
- acknowledges only the activity revision captured by a successful collection
- preserves the latest activity revision when ticks are coalesced
- retains pending activity after a failed collection or when new activity arrives in flight

The hourly reconciliation recovers WSL-only usage, missed native events, client directories created after startup, and cross-day history/limit state that otherwise has no host-side event.

## Compatibility

- live mode keeps polling and event-triggered collection
- fixed intervals keep scheduled collection and disable file watching
- Smart mode is exposed through the existing Collection frequency setting
- settings copy is updated in every bundled locale
- upload interval synchronization from current `main` is preserved
- the limits reset-boundary refresh introduced by #212 remains independent of the Smart activity gate

## Regression coverage

- idle Smart mode still performs the hourly full reconciliation
- WSL-only usage is refreshed by that fallback
- a client directory created after startup is discovered even when its event is missed
- a coalesced tick carries and acknowledges the latest `activityRevision` without a redundant next scan

## Validation

- `npm ci`
- Node 22.23.1: full ESLint and test suite - 1,518 passed, 0 failed, 1 credential-gated integration skipped
- Node 24: `npm run verify` - 1,518 passed, 0 failed, 1 credential-gated integration skipped
- combined Smart, limits reset, settings, and provider regression set - 223 passed
- macOS native-watcher runtime smoke: 3 startup period scans, no immediate scan after a file event, 1 anchored scan at the next tick, and no repeat on the following clean tick
- `git diff --check upstream/main...HEAD`

Rebased onto `upstream/main` at `2edd0a1`, including `fix(limits): refresh expired reset windows (#212)`.


Javis603:
Thanks @wetlink for adding the Smart collection mode. Slotting it into the existing Collection frequency selector is the right shape, and using native filesystem events to cut idle polling is a useful direction.

No rush since it's still a draft, but the main blocker to sort out before it's review-ready is that the activity gate runs before `fullScanDue` is evaluated. When Smart mode has no detected host-side activity, it skips every interval, including the existing hourly reconciliation scan.

This makes native filesystem events the sole source of truth. I previously kept polling enabled after seeing occasional missed events across platforms, and Chokidar cannot recover an event that the underlying OS never reports. It also means:

- WSL-only activity will not refresh because WSL is not watched.
- Client directories created after collector startup may remain undiscovered.
- Missed watcher events have no periodic recovery path.
- Cross-day history, limits, and persisted anchors may remain stale until another host-side event is detected.

Native events are fine as an optimization, but Smart mode still needs an unconditional reconciliation fallback. Please compute `fullScanDue` before the activity gate and skip the interval only when there is no pending activity and no full scan is due.

There is also a smaller coalescing issue: `runTick()` preserves `forceLimits` and `forceHistory`, but drops the latest `activityRevision`. A coalesced scan can process activity without acknowledging it, causing an unnecessary scan on the following interval.

Please add regression coverage for:

- an idle Smart collector still performing the hourly full scan
- WSL-only usage being refreshed by the fallback
- usage from a missed event or a client directory created after startup being recovered by the hourly reconciliation
- the latest `activityRevision` surviving coalescing

The branch is currently conflicting with `main` in `src/electron/main.js` and `src/electron/renderer/index.html`, so it will also need to be rebased before another review.

wetlink:
Thanks for the detailed review. I addressed all requested blockers in `6da1c05` and rebased the branch onto `32533a3` (`v0.32.0`).

- `fullScanDue` is now computed before the activity gate; Smart mode skips only when there is no pending activity and no full scan is due.
- The hourly reconciliation is therefore unconditional and covers WSL-only usage, missed events, client directories created after startup, and cross-day state.
- `runTick()` now retains the highest finite `activityRevision` across coalesced ticks, so a successful scan acknowledges the latest processed activity without causing a redundant next scan.
- Added the four requested regressions: idle hourly fallback, WSL-only refresh, missed/new-directory recovery, and coalesced revision acknowledgement.
- Resolved the `main.js` and `index.html` rebase conflicts while preserving the current upload-interval synchronization and settings layout.

Validation completed on both Node 22 and Node 24: `npm run verify` reports 1,505 passed, 0 failed, and 1 credential-gated integration skipped with an empty `GROK_HOME`. The focused collector/settings/anchor/WSL set reports 86 passed. A macOS native-watcher smoke test also confirmed no immediate event scan, one collection at the next cadence tick, and no repeat on the following clean tick.

The PR is now mergeable, so I am marking it ready for review.

wetlink:
Rebased onto current `main` at `2edd0a1` after #212 and force-pushed the updated wetlink branch (`23676a2`).

The only conflict marker was in the collector state declarations. The resolution keeps both #212's `attemptedResetBoundaries` state/timer path and Smart mode's activity revision state. A range-diff confirms the settings and reconciliation commits are patch-equivalent to the previously reviewed versions; the collector commit differs only by the new upstream limits-reset context.

Post-rebase validation:

- Node 22.23.1: full ESLint + tests, 1,518 passed / 0 failed / 1 credential-gated skipped
- Node 24: `npm run verify`, 1,518 passed / 0 failed / 1 credential-gated skipped
- combined Smart, limits-reset, settings, and provider regression set: 223 passed

The new head will require maintainer approval for the fork CI workflow again.

Javis603:
Thanks @wetlink for the Smart collection work and for following through on the earlier reconciliation feedback.

I pushed one follow-up commit (01d456f1) before merging. It makes successful manual and coalesced scans acknowledge the activity they already covered, and refreshes watcher coverage after full scans so client directories created after startup return to the 10-minute Smart cadence instead of remaining on hourly reconciliation.

Native filesystem events remain an optimization, while the hourly scan stays the correctness fallback. Thanks again for the careful revisions and cross-platform validation.

PR #175, opened 2026-07-17
fix(grok): proxy env + reject deprecated monthly Build fallback

SDS-CC:
## Summary

- Keep Grok CLI `x.ai/billing` as the primary source, then fall back to grok.com's official `GetGrokCreditsConfig` gRPC-web endpoint when the CLI method is unavailable.
- Honor standard `http_proxy` / `https_proxy` / `no_proxy` environment variables (plus uppercase variants and `ALL_PROXY`) for the Grok web request through undici's `EnvHttpProxyAgent`.
- Remove the deprecated `cli-chat-proxy` monthly Build fallback so the displayed meter cannot switch between different quota products.
- Parse gRPC status headers and trailers so rejected credentials are surfaced as re-authentication errors instead of generic availability failures.

## Behavior

No configuration is required for users with direct access to grok.com. Users whose network requires a proxy can launch Token Monitor with the standard proxy environment variables; `NO_PROXY` remains honored. Invalid configured proxy URLs fail closed instead of silently bypassing the proxy.

`undici` is declared as a direct runtime dependency because this code imports `EnvHttpProxyAgent` in production.

## Verification

- `npm run verify` (1444 passed, 1 skipped)
- `npm ci --omit=dev`, then confirmed `undici@7.25.0` and `EnvHttpProxyAgent` are available
- Focused coverage for proxy precedence, `NO_PROXY`, `ALL_PROXY`, quoted values, invalid-proxy fail-closed behavior, retry classification, `GROK_HOME`, and gRPC authentication statuses
- CI passes on Node 22/24 across Ubuntu, macOS, and Windows


Javis603:
Thanks @SDS-CC for adding proxy-aware Grok quota fetching and for identifying the mismatch between the unified credits and deprecated monthly Build quota.

I pushed a follow-up commit (6e67e148) to use undici's `EnvHttpProxyAgent` with `NO_PROXY` and standard environment precedence, ship `undici` as a runtime dependency, remove the deprecated `cli-chat-proxy` fallback, and correctly classify gRPC authentication failures. I also brought the branch up to date with `main`.

PR #185, opened 2026-07-17
fix(history): fill missing Claude Code days from token_stats.db fallback

### LINKED ISSUE #177: 历史额度数量丢失
Image

Image

这里昨天统计的时候是正常的，但是今天看昨天的就是 0 了，而且我中间 345 月有很大面积的为 0

昨天应该是有 4 亿多的用量，今天不知道为啥丢了，刷新了也不行

Code-suphub:
### Bug fix: restore April–May 2026 Claude Code usage

tokscale only reads per-project transcript files (`~/.claude/projects/*/*.jsonl`) for Claude Code usage. During April–May 2026, those transcript files were missing, so the trends dashboard showed 0 tokens for those months — even though the user had ~7 billion tokens of actual usage.

The fix queries `~/.claude/token-stats/token_stats.db` (Claude Code's own SQLite aggregate) as a fallback data source. For each day where tokscale returns zero Claude tokens, the stats DB is consulted. Days where tokscale already has Claude data are never supplemented, avoiding double-counting.

This restored:
- April 2026: 22 active days / 4.5B tokens (was 0)
- May 2026: 23 active days / 2.5B tokens (was 93M from Codex only)
- Active days overall: 63 → 122

### Feature: Tokens/Cost heatmap toggle

The activity heatmap previously always used cost-based intensity coloring. Since `token_stats.db` has no cost data, restored April/May entries would appear blank.

- `computeIntensities` now produces both `intensity` (token-based) and `costIntensity` (cost-based) per day
- A Tokens/Cost segmented toggle is added to both the Dashboard window and the home widget activity module
- The preference is persisted as `heatmapMetric` in settings.json and synced between windows via settings push
- To avoid double-counting, the toggle also passes `intensityKey` through to `contribHeatmap` in usageCharts

## Changes

| Commit | Files |
|--------|-------|
| `fix(history)` | `src/shared/claudeTokenStats.js` (new), `collector.js`, `history.js`, `worker/src/shared/history.js`, `tests/` |
| `feat(dashboard)` | `src/electron/main.js`, `renderer/{dashboard,app,usageCharts}.js`, `dashboard.{html,css}`, `styles.css` |

## Test plan

- [x] `npm run verify` — 1368 tests pass (lint + test)
- [x] `npm run sync:worker` — worker vendored code synced, CI passes
- [x] Manual: April–May 2026 now show Claude Code usage, active days 63 → 122
- [x] Manual: Tokens/Cost toggle works in both Dashboard window and home widget
- [x] Manual: switching metrics does not change active days count
- [x] Manual: toggle persisted across restart

Fixes #177


Javis603:
Thanks @Code-suphub for tracing both parts of this. Your conclusion in #177 correctly identified that manual refresh did not trigger a history rescan, and that the older grey cells were caused by missing Claude Code transcripts.

I opened #187 for the refresh and cross-midnight path. It also fixes two related local bugs: Home held the same full-history snapshot for the entire renderer session, and the activity views used a UTC day key while patching local-day totals. #188 documents Claude Code's default 30-day transcript retention and how to extend it before older daily history disappears.

Before I can properly evaluate the fallback in this PR, I need to understand where `token_stats.db` comes from. On Claude Code 2.1.212, my installation has no `~/.claude/token-stats/` directory and no SQLite database under `~/.claude`, so this path would silently do nothing.

Could you share:

- your Claude Code version
- whether `token_stats.db` appeared automatically or required another tool or setting
- approximately when it was first created

If the database is optional or version-specific, the implementation and documentation should describe it as an opportunistic fallback rather than a generally available Claude Code source.

I would also like the Tokens/Cost heatmap toggle split into a separate PR. The database fallback and its deduplication are a data-source change, while the toggle adds UI, intensity semantics, and a persisted `heatmapMetric` setting. They are related because the database has no cost data, but I want to review those two compatibility decisions independently.

Code-suphub:
Thanks for the detailed review!

**`token_stats.db` origin**

- Claude Code version: 2.1.149
- The DB appeared automatically — no additional tool or setting was needed. It was first created on 2026-04-24 and contains records from 2026-03-30 onwards (120k+ token rows across ~110 days).
- That said, you noted that your installation (Claude Code 2.1.212) has no `~/.claude/token-stats/` directory at all. This suggests the DB may have been removed or made optional in a later version. The fallback is already designed as opportunistic (silently returns null when the DB is missing), so it's safe either way — but I'll update the implementation to make this more explicit.

**Heatmap toggle split**

Per your request, the Tokens/Cost heatmap toggle has been split out into a separate PR: #190. This PR now contains only the history fix commits:

1. `fix(history): fill missing Claude Code days from token_stats.db fallback`
2. `fix(history): dedup Claude Stats at graph level, add costIntensity tests`
3. `fix(collector): dedup Claude Stats against raw tokscale contributions`

Code-suphub:
Resolved merge conflicts with #193 (daily history archive). Both approaches are now integrated:

1. **PR #193's archive** (`retainDailyHistory` / `dailyHistoryArchive.js`) — persists every observed daily observation so future source cleanup doesn't erase it. This is the primary protection going forward.

2. **`token_stats.db` fallback** — retained as a supplementary source for Claude Code data that was deleted before Token Monitor first observed it (e.g. the April–May 2026 transcripts). It only fills days where tokscale reports zero Claude Code tokens.

The two layers work together: the archive captures data from this point forward, while the DB fallback helps recover past data the archive never saw.

Key merge details:
- `collector.js`: both `retainDailyHistory` and `buildClaudeStatsGraph` run in sequence — archive first, then DB fallback if archive is disabled or fails
- `main.js` / `agent.js`: updated with archive wiring from #193
- `history.js`: `computeIntensities` now produces both token-based `intensity` and cost-based `costIntensity` (required by the heatmap toggle in #190)
- New files: `dailyHistoryArchive.js`, `dailyHistoryArchive.test.js`, `dailyHistoryArchiveWiring.test.js`

Javis603:
Thanks @Code-suphub for clarifying the version and for splitting the heatmap toggle into #190.

I checked the published Claude Code 2.1.149 macOS package corresponding to my environment. It contains the built-in `stats-cache.json` implementation, but I found no reference to `~/.claude/token-stats/`, `token_stats.db`, or the SQL schema used here. The same is true of the current 2.1.215 release, and I could not find this database documented by Anthropic.

Because of that, I can't verify this as a Claude Code-owned data source or rely on its schema as a compatibility surface. Returning `null` when the file is absent avoids a crash, but Token Monitor would still be carrying support for an undocumented database whose producer and semantics are unknown.

The daily history archive now on `main` already protects newly observed data going forward. Without a verifiable producer and schema contract, I don't think this fallback should be part of Token Monitor, so I'm going to close this PR without merging.

Thank you for tracing the missing-history case and for the work you put into the implementation.

PR #218, opened 2026-07-20
feat: add Trae Work (TRAE SOLO CN) token usage collection

talesov:
## Summary

Add support for collecting token usage data from **Trae Work** (TRAE SOLO CN), a Rust-based AI coding assistant that stores session data in a SQLCipher 4 encrypted database.

## How it works

Trae Work stores its data in an encrypted SQLCipher database at:
- Windows: `%APPDATA%\TRAE SOLO CN\ModularData\ai-agent\database.db`

Since the database is encrypted, users need to provide the raw key (64-char hex) via one of:
1. Environment variable: `TOKEN_MONITOR_TRAE_KEY`
2. Key file: `trae-key.json` in the app data directory

A helper script `scripts/extract-trae-key.py` is included to extract the key from process memory (Windows, requires admin).

## Changes

- **`src/shared/traeUsage.js`** (new): Core module - resolves key, reads SQLCipher DB, builds tokscale-compatible JSON
- **`scripts/extract-trae-key.py`** (new): Windows memory scanner to extract SQLCipher key from ai_agent.dll process
- **`src/shared/collector.js`**: Integrate trae collection (watch paths, period merging, history graph)
- **`src/shared/clientTracking.js`**: Add `trae` to DEFAULT_CLIENTS
- **`src/shared/usage.js`**: Add trae normalization rule
- **`src/electron/renderer/app.js`**: Add trae to UI labels and known clients
- **`src/electron/renderer/usageCharts.js`**: Add trae chart color
- **`src/electron/discordRpc.js`**: Add trae to known client assets
- **`.env.example`**: Document TOKEN_MONITOR_TRAE_KEY
- **`tests/shared/clientTracking.test.js`**: Include trae in locally parsed clients

## Database schema notes

- `history_v2` table: `token_usage` column contains a plain integer (output tokens per turn)
- Model info is in `messages` JSON column: `raw_messages[].extra_info.model`
- `server_history_info` table: session metadata (title, timestamps)

## Testing

Tested locally against a real Trae Work database (~615MB, 18k+ records, 101 sessions, 26 models).

## Notes

- Uses `better-sqlite3` with SQLCipher extension if available, falls back to `sqlcipher` CLI
- No keys or sensitive data are hardcoded; all credentials come from user configuration

---
## Summary by cubic
Add Trae Work (TRAE SOLO CN) usage collection by decrypting its SQLCipher DB and merging results into totals and history. Improves security (stdin key, atomic key writes) and performance (single-pass DB read), with full UI integration.

- **New Features**
  - `src/shared/traeUsage.js`: reads encrypted DB via `better-sqlite3` (SQLCipher) or `sqlcipher` CLI; feeds key via stdin; single-pass decrypt/query; emits `tokscale`-compatible periods and history; warns if no SQLCipher backend.
  - Collector: filters `proma`/`trae` from `tokscale` calls; shares Trae rows between periods and history; adds watch paths; merges usage into today/month/all-time.
  - UI: adds Trae labels, chart color, theme vendor order/labels, and Discord label/asset.
  - Config/tests/worker: `.env.example` documents `TOKEN_MONITOR_TRAE_KEY`; adds `trae` to default/known clients; tests updated; `normalizeClientName` recognizes `trae`.

- **Migration**
  - Provide the 64-char key via `TOKEN_MONITOR_TRAE_KEY` or a `trae-key.json` file in the shared data dir or next to the DB.
  - On Windows, run `python scripts/extract-trae-key.py` (admin, Trae running). The script now validates the key by opening the DB and writes `trae-key.json` atomically.

<sup>Written for commit 83bbb00ac688e5f4aa3ef8fbfa1c2e03efe62f37. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/218?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,8 +23,13 @@ TOKEN_MONITOR_SYNC_UPLOAD_INTERVAL_MS=0
 
 # Comma-separated client list. Optional — leave unset to track all supported tools.
 # Set to empty to disable tracking.
-# Example: claude,codex,hermes,opencode,openclaw,cursor,antigravity,cline,kimi,qwen,grok,copilot,pi,zed,kilocode,micode,zcode,kiro,codebuddy,workbuddy,proma
-TOKEN_MONITOR_CLIENTS=claude,codex,hermes,opencode,openclaw,cursor,antigravity,cline,kimi,qwen,grok,copilot,pi,zed,kilocode,micode,zcode,kiro,codebuddy,workbuddy,proma
+# Example: claude,codex,hermes,opencode,openclaw,cursor,antigravity,cline,kimi,qwen,grok,copilot,pi,zed,kilocode,micode,zcode,kiro,codebuddy,workbuddy,proma,trae
+TOKEN_MONITOR_CLIENTS=claude,codex,hermes,opencode,openclaw,cursor,antigravity,cline,kimi,qwen,grok,copilot,pi,zed,kilocode,micode,zcode,kiro,codebuddy,workbuddy,proma,trae
+
+# Trae Work — SQLCipher database key (64-char hex) for reading local usage.
+# Extract with: python scripts/extract-trae-key.py (Windows, admin, Trae running)
+# Or place a trae-key.json ({"key":"<64hex>"}) in the Token Monitor data dir.
+TOKEN_MONITOR_TRAE_KEY=
 
 # Projects — collect local workspace metadata for the Projects view.
 # Optional — defaults to disabled. Set to 1 to enable.
--- a/scripts/extract-trae-key.py
+++ b/scripts/extract-trae-key.py
@@ -179,8 +179,62 @@ def scan_memory_for_key(pid, db_path):
         if _verify_key(key, page1):
             return key
 
-    # If HMAC verification fails, return first candidate (try direct open)
-    return candidates[0] if candidates else None
+    # HMAC didn't match (SQLCipher variants differ) — prove the key works by
+    # actually opening the database before claiming success.
+    for key in candidates:
+        verdict = _validate_key_opens_db(key, db_path)
+        if verdict is True:
+            return key
+
+    # No candidate could be validated. Only fall back to the first candidate
+    # when no validation backend exists at all, with a loud warning.
+    if candidates:
+        if _validate_key_opens_db(None, None) is None:
+            print("[!] WARNING: could not verify key (HMAC mismatch and no sqlcipher3 module).")
+            print("    Saving first candidate anyway — collection may fail if it is wrong.")
+            return candidates[0]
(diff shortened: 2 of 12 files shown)


Javis603:
Thanks @talesov for working on Trae CN support. I’ve completed a full review of the integration. I understand why this approach reads the local encrypted database instead of relying on an API. Local CN support could be useful, but I don’t think the current implementation is ready to merge.

The first issue is the client boundary. Token Monitor does not currently expose Trae as a tracked client, but its bundled Tokscale already uses `trae` for the international Trae IDE and Trae Solo integration. Reusing that same ID for the CN database parser would create a client-ID and data-source collision, making the international integration unnecessarily disruptive later. The local parser should therefore use a distinct, default-off ID such as `trae-cn`, leaving `trae` available for the future Tokscale-backed international client.

The current separation between the application and the key-extraction helper is appropriate. The Electron application does not invoke `scripts/extract-trae-key.py` automatically, request administrator privileges, or read Trae process memory. The runtime collector only consumes a key that has already been supplied through the environment or a recognized local file.

Keeping the helper in the repository for advanced users is acceptable, provided that it remains strictly manual and is excluded from all packaged builds. Its documentation should clearly state that it is Windows-only, requires elevated access, may trigger antivirus or EDR detection, and may stop working when Trae changes.

The helper also needs to fail closed. It must never save an unverified candidate, overwrite an existing working key after validation failure, or print the raw key to the terminal. Any persisted key must be treated as sensitive and stored with appropriate local file protections.

The remaining implementation requirements are:

- Use a separate, default-off `trae-cn` client ID.
- Detect and report `missing key`, `missing SQLCipher backend`, and `invalid key` states instead of silently returning empty usage.
- Add sanitized fixtures and direct tests for every claimed CN variant, covering input, output, cache, model, period, and history totals. Please also document which variants were actually verified.
- Avoid reparsing the full history database during watch-triggered refreshes.
- Complete the standard client integration, including icon assets, `clientsWithIcon`, row-icon CSS, README translations, advanced setup documentation, and collector tests.
- Clearly document that this is an advanced local integration with user-managed prerequisites and that it may break if Trae changes its encrypted database schema.

The branch now also conflicts with current `main`, so it will need to be rebased after the design changes. Once the client split, advanced setup contract, status reporting, and verified token mapping are addressed, I’m happy to re-review.

PR #239, opened 2026-07-23
feat(updater): add automatic update downloads

CxHsin:
## Summary

- add an opt-in Automatic updates switch to App Updates settings
- automatically download a newer release after startup, scheduled, or manual checks when enabled
- keep the existing restart-to-install prompt so updates never restart the app unexpectedly
- localize the setting in every bundled language and cover the settings/update wiring with tests

## User impact

Users who enable Automatic updates no longer need to click through the separate check and download steps. The app prepares the update in the background and asks the user to restart once it is ready.

This is a refreshed submission of #202, rebased onto the latest `main`.

## Screenshot

Automatic updates setting

## Validation

- `node --test tests/electron/appUpdateState.test.js tests/electron/appUpdateReleaseNotes.test.js tests/electron/i18n.test.js` (30 passed)
- `npm run verify` (lint passed; 1,686 tests passed, 2 skipped, and 2 existing Windows path-sensitive assertions failed in `tests/shared/collectorLoadGuards.test.js` for local Copilot/Hermes directories)

---
## Summary by cubic
Adds an opt‑in Automatic updates setting that downloads new releases in the background and refines the update pill for clearer, safer installs. The app prompts to restart when ready and never restarts on its own.

- **New Features**
  - Added “Automatic updates” toggle under App Updates.
  - When enabled, new versions auto‑download after startup, scheduled, and manual checks.
  - Keeps the existing restart‑to‑install flow; no automatic installs.
  - Setting is persisted and localized.

- **Bug Fixes**
  - Enabling Automatic updates triggers an immediate update check that bypasses the background‑check cooldown.
  - Automatic downloads skip versions the user dismissed; the toggle disables with a clear reason on unsupported or busy builds, and the footer update pill adds a dedicated Restart button while keeping release notes on the main action with improved accessibility.
  - Added tests to ensure automatic downloads don’t start when the setting is off and to cover the full automatic‑download state matrix.

<sup>Written for commit 0cbd5e689fe218269c3c5ce0f6bf5d54e18503e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/239?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Javis603:
Thanks @CxHsin for refreshing this PR. Keeping automatic downloads opt-in while requiring an explicit restart to install fits the updater flow well.

I pushed a follow-up commit (7b2aa879) to finish the UX and edge cases: unsupported builds now disable the setting with a clear reason, the copy accurately describes automatic downloading, the ready state keeps release notes accessible alongside a dedicated Restart action, and the download decisions now have behavior-based matrix tests.

PR #266, opened 2026-07-26
feat(limits): show balances as money and let them drive the tray

Javis603:
## What changed

A balance was rendered three different ways depending on where you looked at it:

| Surface | Before | After |
|---|---|---|
| Limits page | `$25.00` | unchanged |
| Home | `25% left` | `$25.00` |
| Tray | DeepSeek not selectable at all | selectable, shows `¥4.00` |

The percentage on Home was real arithmetic, but the denominator was not a quota. For a third-party relay account the window's `used` is the provider's **cumulative** counter — `used_quota` / `total_used`, surfaced as `allTimeSpend` — so `limit` becomes `remaining + everything ever spent`. An account with `$25` left after `$75` of cumulative usage showed `25%`, from `25 / (25 + 75)`. That number falls the longer you use the account and jumps back up every time you top up, so it reads like a quota while measuring something else entirely.

(This is a different denominator from the one further down under *Percentages stay off the wire*, which covers the derived meter for balances that report no percentage at all.)

## Why it happened

Money had no marker in the shared limit model — only `usedPercent` / `remainingPercent` were comparable. So each surface improvised:

- The **limits page** looked right because it has a hand-written branch per provider (`openrouter` / `thirdparty` / `deepseek` / `mimo`) that pushes the amount in as a value override.
- **Home** is generic. It only printed money for a synthesized `kind: 'balance'` window, and that window was built behind a hardcoded `providerId === 'deepseek' || providerId === 'mimo'` whitelist. Everything outside the whitelist fell through to the percentage path.
- The **tray** filtered to windows with a non-null percentage, and DeepSeek ships no windows at all, so it could never be picked.

## The fix

`windows[].metric === 'credits'` already existed in the wire shape and was already sent by the OpenRouter and third-party collectors. This extends that convention rather than inventing a mechanism:

- `normalizeLimitWindow` gained a `currency` field, so a balance renders in its own currency with no conversion (DeepSeek is CNY, the rest USD).
- The DeepSeek and MiMo collectors now emit a `credits` window too, so `windows` is no longer empty for them.
- Home, the tray and the limits page all key off the marker through one new shared module, `src/shared/limitBalanceDisplay.js`. The `deepseek || mimo` whitelist and the duplicated meter derivation are gone.

Two things stayed deliberately unchanged: **no new tray style** (the existing percentage styles print compact money when the selected window is a balance, so `STYLE_GROUPS`, previews, i18n and saved layouts are all untouched), and **the limits page renders identically** — only its private copy of the meter math was replaced with the shared one.

## Selections carry the resolved percentage

Letting a balance-only provider reach the tray exposed a second instance of the same bug class this PR is about. Three consumers re-derived their percentage from the raw window instead of reading what the picker had already resolved. A balance window has no wire percentage, so re-deriving turned 40% remaining into a fabricated `0%` and an empty, exhausted-looking bar:

| Tray surface | Symptom with a balance pick |
|---|---|
| single-provider bar icon | empty bar |
| all-sessions icon | printed `0%` |
| two-provider bar icon | empty bar |

`compactLimitSelection` now exposes `primaryPercent` / `secondaryPercent`, and every icon renderer reads those or the picker's `remaining`. Sweeping the remaining direct reads of `remainingPercent` / `usedPercent` confirmed the other eleven are safe: each sits inside a credits-aware branch, consumes an already-mapped window, or belongs to Antigravity, which has no balance windows.

## Percentages stay off the wire

Balances that report no percentage of their own still need something to fill a meter with. That number is derived from **observed month spend** (`amount / (amount + monthSpend)`) — distinct from the cumulative denominator above — and it is a visual hint, not a fact. It is derived in the display layer and never written to the wire, so the hub and the Worker never store an invented number.

That choice also turned out to carry backward compatibility for free: an older widget receiving a new record drops the percentage-less window through its own existing filter and keeps synthesizing its own row, so nothing is duplicated.

## Mixed-version fleets

The reverse direction did need a shim. A device older than this change posts a balance with **no** window, which every new renderer would drop — silently removing the row from Home. Updates in this app are fully manual (`autoDownload` and `autoInstallOnAppQuit` are both `false`) and the headless agent has no updater at all, so mixed-version fleets persist indefinitely rather than for a brief window.

`normalizeLimitProvider` therefore restores the window on ingest. That is the one funnel both the local collector output and hub aggregation pass through, so no surface has to remember to call it. Only the amount is restored, never a percentage. Verified across all four combinations:

| | Before | After |
|---|---|---|
| old record → new Home | **(empty)** | `Balance` with its amount |
| old record → new tray picker | **(empty)** | `Balance` |
| new record → new Home | 1 row | 1 row |
| new record → old Home | 1 row | 1 row |
| third-party record → new Home | 1 row | 1 row (no-op) |

## Testing

`npm run verify` — 1886 passing, lint clean. New coverage: `currency` normalization, the restored-window compatibility path (including no-duplicate and percentage-free assertions), DeepSeek and MiMo collector output, Home rendering money for credits windows, MiMo's plan and balance sitting side by side, Home sorting a nearly drained balance ahead of a healthy quota, and the tray picker accepting a percentage-less credits window, plus a regression test pinning that a balance selection exposes resolved percentages rather than leaving consumers to re-derive them.

Three tests that asserted against source-code strings sliced out of `app.js` now require the shared module and assert behaviour instead.

## Known follow-ups, not in this PR

- The limits page still carries four hand-written per-provider branches that duplicate what the shared module now knows.
- `trayLayout.js` and `trayText.js` still hold two parallel implementations of `meteredWindows` / `preferredWindow`, and their signatures now differ (`windowPercent(provider, window)` vs `remainingPercent(window, provider)`). Worth unifying in a separate change.
- The compatibility shim in `normalizeLimitProvider` has no version anchor; it is removable once no supported device predates this change.


Javis603:
## Review follow-up

### Fixed: an empty balance rendered as a full meter (aa63ff3b)

Correct catch, and the worse half of the bug class this PR is about — the earlier tray defect fabricated an alarming `0%`, this one fabricated a reassuring full bar, which nobody would think to check.

A zero balance with no observed month spend fell through to the "untouched, so full" fallback. A freshly tracked exhausted account showed its real `$0.00` next to a full meter and sorted ahead of genuinely healthy quotas on Home.

| Balance state | Before | After |
|---|---|---|
| exhausted, no spend observed yet `(0, 0)` | **100%** | `0%` |
| **overdrawn** `(-5, 0)` | **100%** | `0%` |
| exhausted, spend observed `(0, 12)` | `0%` | unchanged |
| untouched positive balance `(4, 0)` | `100%` | unchanged |
| in normal use `(4, 6)` | `40%` | unchanged |

The overdrawn row was not flagged but fails the same way, so it is covered too. Zero funds is zero remaining regardless of what spend has been observed; the full-meter fallback now applies only to a positive balance with no spend yet.

Worth recording that this behaviour predates the branch — it came over unchanged from the limits page's own copy of the derivation (`total > 0 ? … : 100`). Folding that copy into the shared module is what widened its reach from a single meter to Home ordering, Home colouring and the tray icons, so it is fair to land the fix here.

### Not changing: the "lifetime spend" wording

This one is a misread, and the distinction matters enough to keep the wording as it is.

The sentence describes the **third-party / OpenRouter** path, and that denominator really is cumulative: the window's `used` comes from `used_quota` / `total_used` (`thirdPartyLimits.js:265,276`), stored as `allTimeSpend`.

The `monthSpend` denominator belongs to a different path — the old synthesized `balanceWindow()` — and that path never printed a percentage at all:

```js
// app.js:2379, before this branch
if (window?.kind === 'balance') {
  return formatMoney(window.amount, window.currency);   // money, not a percentage
}
```

Its `monthSpend`-based percentage only fed the bar and the sort order. Every percentage a user actually saw on Home came from a real third-party window with a lifetime denominator.

The description already separates the two: "lifetime spend" for the relay case, and `amount / (amount + monthSpend)` where it describes the derived meter.

Javis603:
## Description clarified

Not switching the word to `monthSpend` — that would make the sentence wrong — but the description now names the code path and the source fields outright, because two denominators live in this PR and the text was leaning on context to tell them apart:

| Where | Denominator | Source |
|---|---|---|
| the percentage Home used to show for relay accounts | cumulative account usage | window `used` ← `used_quota` / `total_used`, surfaced as `allTimeSpend` (`thirdPartyLimits.js:213,265,276`) |
| the derived meter for balances that report no percentage | observed month spend | `provider.balance.monthSpend` (`limitBalanceDisplay.js`) |

The first is what the "What changed" section describes: it is the number users actually saw as `% left`, and it really is cumulative. The second is the new derivation, and it is now labelled "observed month spend" in its own section, with a pointer between the two so neither can be read as the other.

A careful reader landing on the wrong one twice is fair evidence the prose was doing too much work, even where it was accurate — so the fix is precision rather than a correction.

PR #268, opened 2026-07-27
fix(limits): clear stale provider status on settings change

SunRicardo:
## 概要

修复更改 Provider API Key 后设置页面仍显示「无法获取」的问题。

当用户在设置中修改某个 Provider 的 API Key（如 DeepSeek）时，`settings:update` 会刷新对应的 limits 探针。旧实现没有先清除该 provider 已发布的状态，因此新探针完成前，设置页面仍可能渲染上一次的结果。

### 具体改动

- `src/electron/main.js`：所有 settings-change limits invalidation 都先清除受影响的 provider/account lane，再启动新探针；Runtime 结构性重启分支也会保留同样的 clear-before-refresh 语义
- `src/electron/deviceRuntimeCoordinator.js`：抽出纯数据 invalidation plan 和可测试的执行器，并统一使用 options object，避免含义不明的位置布尔参数
- `src/electron/renderer/app.js`：移除固定 2.5 秒延迟和第二次全量刷新，探针完成后由现有 stats push 更新界面，避免重复触发完整 usage tick 和 provider-wide limits refresh
- 测试：以 fake runtime 行为测试覆盖 settings-change plan、clear-before-refresh 调用顺序和 clear-only 路径，取代读取源码并用 regex 匹配实现格式的测试

### 行为取舍

Credential identity 改变时会刻意丢弃旧 credential 的 retained last-good snapshot。因此，如果更换 credential 后的第一次探针失败，limits 可能暂时不显示，而不会继续展示属于旧 credential 的数据。

### 问题复现

1. 输入一个有效的 DeepSeek API Key，保存
2. 点击「清除密钥」
3. 重新输入同一个 Key，保存
4. 设置页面显示「无法获取」，但主页额度显示正常

### 根因与修复

`settings:update` 原本只刷新对应 lane，没有先清除已发布的旧状态：

```js
void queueLimitInvalidation(scope, 'settings-change');
```

现在 settings-change plan 会为每个受影响的 scope 明确携带 `{ clear: true }`，再由共用执行器按 clear-before-refresh 顺序处理：

```js
return scopes.map((scope) => ({
  scope,
  reason: 'settings-change',
  options: { clear: true }
}));
```

固定延迟重试并不是修复的一部分。Limits completion 已经会通过现有 stats push 更新 renderer，而再次调用强制 refresh 会重复执行完整 usage 和 limits 工作。

### 测试

```bash
npm run verify
```

1902 tests：1901 pass / 0 fail / 1 skipped。

---

<details>
<summary>English version</summary>

## Summary

Fix the settings page showing "unavailable" after changing a provider API key.

When a user modifies a provider API key such as DeepSeek, `settings:update` refreshes the affected limits probe. Previously it did not first clear the provider's published status, so the settings page could continue rendering the previous result until the new probe completed.

### Changes

- `src/electron/main.js`: Clear every affected provider/account lane before refreshing it after a settings change. Structural runtime restart branches preserve the same clear-before-refresh behavior.
- `src/electron/deviceRuntimeCoordinator.js`: Extract a pure invalidation plan and a testable executor, and replace positional booleans with an options object.
- `src/electron/renderer/app.js`: Remove the fixed 2.5-second delay and second full refresh. Existing stats pushes update the UI when the probe completes, avoiding a duplicate full usage tick and provider-wide limits refresh.
- Tests: Replace source-regex assertions with fake-runtime behavior tests for the settings-change plan, clear-before-refresh call order, and the clear-only path.

### Behavior tradeoff

Credential identity changes intentionally discard the retained last-good snapshot from the old credential. If the first probe after a credential change fails, limits may temporarily disappear rather than continue displaying data that belongs to the previous credential.

### Reproduction

1. Enter a valid DeepSeek API key and save.
2. Click "Clear key."
3. Re-enter the same key and save.
4. The settings page shows "unavailable" while the home page displays the balance correctly.

### Root cause and fix

`settings:update` previously refreshed the affected lane without first clearing its published status:

```js
void queueLimitInvalidation(scope, 'settings-change');
```

The settings-change plan now gives every affected scope an explicit `{ clear: true }` option, which the shared executor applies in clear-before-refresh order:

```js
return scopes.map((scope) => ({
  scope,
  reason: 'settings-change',
  options: { clear: true }
}));
```

The fixed delayed retry is not part of the fix. Limits completion already reaches the renderer through existing stats pushes, while another forced refresh repeats full usage and limits work.

### Testing

```bash
npm run verify
```

1902 tests: 1901 passed / 0 failed / 1 skipped.

</details>


Javis603:
Thanks @SunRicardo for fixing the stale provider state after credential changes. The { clear: true } change matches the existing credential lifecycle behavior and remains the core fix.

I pushed a follow-up commit (fb43b646) to remove the fixed 2.5-second retry and second full refresh, since limits completion already reaches the renderer through stats pushes. I also moved settings invalidation behind testable plan and executor helpers, replacing the source-regex tests with fake-runtime behavior tests for the actual clear-before-refresh sequence.

All platform CI checks are green. Thanks for the contribution.

PR #282, opened 2026-07-29
perf: reduce background collection and hidden rendering work

REMIXPLAYER:
## Summary

- use native filesystem events for live collection on macOS while retaining polling on Windows and Linux
- map watch activity to tracked clients and refresh only their `--today` partitions
- retain all-client fallback, hourly full reconciliation, WSL refresh behavior, and full retry after a failed smart scan
- coalesce stats rendering while the Electron window is hidden, then render the latest state when it becomes visible
- preserve visible-window refreshes, immediate state updates, and tray icon updates

## Why

Issue #260 reports unusually high battery impact on macOS. Profiling showed repeated Tokscale scans as the main cost, with renderer work continuing in the background as a smaller contributor.

The collector previously used 2-second polling in macOS live mode and every watch refresh scanned all tracked clients. The renderer also rebuilt stats-driven UI while its window was hidden.

## Compatibility

This does not change settings, APIs, wire data, or visible-window refresh behavior. Windows and Linux live mode continue using polling. Unknown watch paths fall back to an all-client scan, failed smart scans retry all clients on the next active interval, and a full reconciliation still runs at least hourly.

## Verification

- `npm run verify`
- 2026 passed, 0 failed, 1 skipped
- regression coverage for native/polling selection, targeted and fallback collection, failed smart-scan retries, per-client aggregation, WSL reconciliation, and hidden renderer updates

## Energy observation

Activity Monitor's rolling 12-hour energy value decreased from approximately 310 to 290 after the renderer change and to approximately 220 after the combined collector and renderer changes (about 29% cumulatively). This is a preliminary field observation, not a controlled A/B benchmark.

Refs #260


REMIXPLAYER (commented):
This PR avoids unnecessary stats rendering while the window is hidden, then updates it with the latest data when shown again. It also uses native filesystem events for macOS live tracking and refreshes only the affected client, reducing background work without changing the visible UI or usage results.

REMIXPLAYER:
## macOS idle power measurement

I compared this branch with the released **v0.36** macOS arm64 build using `powermetrics`.

- Environment: Apple Silicon, macOS 26.3.2
- Scenario: live tracking enabled, no AI-session file changes, window visible
- Method: three alternating 5-minute samples per build; values below are the median of each run's per-sample median, summed across the main process, Helper, and Renderer.

| Metric | v0.36 | This PR | Change |
| --- | ---: | ---: | ---: |
| CPU | 4.03 ms/s | 0.61 ms/s | -85% |
| Energy Impact | 0.40 | 0.04 | -90% |
| Wakeups | 6.81/s | 3.88/s | -43% |
| Energy Impact P90 | 1.42 | 0.28 | -80% |
| Samples with Energy Impact >= 50 | 6.1% | 2.4% | -61% |

The raw `powermetrics` output was retained locally for all six runs.

This is an idle-background measurement, not a synthetic active-conversation benchmark, so it demonstrates reduced baseline work rather than a guaranteed end-to-end power reduction for every workload.

Javis603:
Thanks @REMIXPLAYER for investigating the polling watcher, and especially for measuring it with `powermetrics` rather than relying on Activity Monitor. The idle results give solid support for moving macOS Live mode to native events.

I pushed a follow-up commit ae8c48f4 to this branch. It fixes one correctness bug in the targeted refresh path: `takeWatchClients()` drains the pending client set before collection, but the failure path only forced a full rescan when `intervalRequiresActivity` was enabled. In Live mode, a failed targeted tick therefore lost its client scope. A later event for another client scanned only that client, leaving the failed partition stale until the 5 to 30 minute interval reconciliation. The follow-up forces the next tick to cover every client in every mode and adds a Live regression test that fails without the fix.

The commit also restores a Token Monitor-owned `TOKEN_MONITOR_WATCH_POLLING` override for network mounts, some FUSE drivers, and container bind mounts that do not reliably deliver native events. Chokidar's `CHOKIDAR_USEPOLLING` does override an explicit `usePolling: false`, but that is Chokidar's configuration surface rather than ours. I also added guards for the client ID and alias invariants that targeted partitions now depend on.

I checked the targeting premise against Tokscale 4.7.0's source. `--client` narrows the enabled client set before scan tasks are built, and client-specific SQLite discovery is behind the same filter. A targeted scan therefore skips unrelated directory walks and database reads rather than scanning everything and filtering afterward.

Because the `powermetrics` scenario is intentionally idle and never exercises targeted scans, I also ran a local scanner benchmark using the watch-tick command shape:

`--json --client <csv> --group-by client,session,model --today`

Each result is the median of seven alternating runs on macOS, on battery with Low Power Mode enabled. Claude was excluded from every variant because its data tree was changing during the measurement.

| Variant | Wall time | CPU, user + system |
| --- | ---: | ---: |
| All clients except Claude | 0.340s | 0.510s |
| `--client codex` | 0.250s | 0.400s |
| `--client cursor` | 0.200s | 0.200s |

In this sample, targeting reduced scanner CPU by roughly 22% for Codex and 61% for Cursor. The selected client's data set still dominates the cost, and `.codex` is 6.6 GB on this machine, so targeting Codex retains much of the full-scan work. The gain is real, but it is secondary to removing the 2-second polling loop, which is the change exercised by the idle benchmark.

REMIXPLAYER:
@Javis603 Thank you for following up on #284 and turning the idea into a robust cross-platform implementation. I appreciate the careful handling of watch-descriptor exhaustion, the explicit polling fallback, and the real filesystem coverage across all platforms. This is a much more complete solution than the initial proposal.Thank you for the thorough review and for catching the failed targeted-refresh case.

PR #287, opened 2026-07-30
feat(mac): prompt to move app to Applications

fizzy718:
## Summary

- prompt packaged macOS builds to move into Applications when launched elsewhere
- use Electron's native move flow so the app is copied and relaunched automatically
- require confirmation before replacing an existing installation and surface move failures
- cover macOS-only gating, confirmation, conflicts, and error handling with tests

## Why

The GitHub release includes a DMG for manual installation and a ZIP required by the automatic updater. Users can bypass the DMG's drag-to-Applications view or mistakenly download the ZIP, then run the app directly from the mounted image or Downloads. Previously the app launched without explaining that it had not been installed.

## User impact

Packaged macOS builds started outside Applications now offer a native **Move to Applications** action. Development builds, other platforms, and already-installed copies are unaffected.

## Validation

- `npm run lint`
- `npm test` — 2045 passed, 1 skipped, 0 failed

## Screenshot

Not attached yet because the native prompt only appears in a signed packaged build running outside Applications; this PR remains draft for release-package smoke testing.

---
## Summary by cubic
Prompt users on macOS to move the packaged app to Applications when launched from elsewhere. Uses the native move flow with conflict confirmation, auto-relaunch, and clear error handling.

- **New Features**
  - Show native “Move to Applications” when a packaged macOS build runs outside Applications.
  - Confirm before replacing an existing install; do not replace if that copy is running.
  - Auto-relaunch after a successful move; show an error message on failure.
  - Tests cover macOS gating, confirmation, conflicts, and errors.

<sup>Written for commit b795e8d13adc790708c9eb7008efa49960b8ba93. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/287?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/electron/macosAppInstall.js
+++ b/src/electron/macosAppInstall.js
@@ -1,0 +1,62 @@
+'use strict';
+
+function shouldOfferMacAppInstall({ app, platform = process.platform } = {}) {
+  return platform === 'darwin'
+    && Boolean(app?.isPackaged)
+    && typeof app.isInApplicationsFolder === 'function'
+    && !app.isInApplicationsFolder();
+}
+
+function replaceExistingApp(dialog, appName, conflictType) {
+  const running = conflictType === 'existsAndRunning';
+  const response = dialog.showMessageBoxSync({
+    type: 'warning',
+    buttons: ['Replace', 'Cancel'],
+    defaultId: 1,
+    cancelId: 1,
+    noLink: true,
+    message: `${appName} is already in Applications`,
+    detail: running
+      ? 'The installed copy is currently open. Close it, then try moving this copy again.'
+      : 'Replace the existing copy with this one?'
+  });
+  return !running && response === 0;
+}
+
+async function offerMacAppInstall({ app, dialog, appName, platform = process.platform } = {}) {
+  if (!shouldOfferMacAppInstall({ app, platform })) return false;
+
+  const { response } = await dialog.showMessageBox({
+    type: 'question',
+    buttons: ['Move to Applications', 'Not now'],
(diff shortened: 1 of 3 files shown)


Javis603:
Thanks @fizzy718 for adding the macOS Applications-folder prompt and covering the packaged-app gating, conflicts, and move failures.

The overall direction is right. Users who drag the app into Applications from the DMG as intended are unaffected, while users who launch it directly from a mounted DMG or run an extracted copy from a ZIP or Downloads get a native path to install it properly.

I found one UX issue to adjust before this is ready: when the existing Applications copy is running, `replaceExistingApp()` still shows a `Replace` button but always returns `false`, so pressing it cannot perform the action the button describes. Please either make this an informational dialog with an appropriate action such as `OK`, or let Electron’s default running-copy behavior handle that conflict instead of showing a misleading `Replace` action.

Before merging, I’ll run a signed packaged macOS smoke test from a mounted DMG and an extracted ZIP, covering decline, successful move and relaunch, and existing closed or running copies. The unit tests cover the decision logic, but not Electron’s native move and relaunch behavior.

Once those are covered, I think this will be in good shape for another review.

PR #345, opened 2026-08-06
fix(tray): size Windows tray icon DPI-aware

### LINKED ISSUE #314: Windows: system tray icon too small / hard to read percentage
## Summary

On **Windows**, the Token Monitor **system tray / notification-area icon is extremely small** compared with neighboring tray icons. The usage percentage (e.g. `23%`) is hard to read at a glance.

## Screenshot

Windows tray: Token Monitor icon is much smaller than adjacent icons (highlighted)

Red box marks the Token Monitor tray glyph. Adjacent icons (language, third-party, network, volume, battery) render at a normal Windows tray size; Token Monitor’s circular percentage icon appears undersized / soft.

## Environment

| Item | Value |
|------|--------|
| **Token Monitor version** | **0.39.0** (matches current GitHub release `v0.39.0`) |
| **Client / distribution** | Official **desktop app** — Electron installer build `Token Monitor.exe` |
| **Install path** | `%LOCALAPPDATA%\Programs\token-monitor\` |
| **Update channel** | GitHub releases / electron-updater (`token-monitor-updater`, publisher SignPath Foundation) |
| **OS** | **Windows 11** Home (Chinese), version **10.0.26200**, **64-bit** |
| **Display** | Primary **1920×1080**, **DPI 96 (100% scale)** — *not* high-DPI scaling |
| **Taskbar** | Windows 11 notification area (system tray), next to IME / network / volume / battery |

## Expected

Tray icon should match standard Windows notification-area icon metrics (typically crisp **16×16** logical, with proper **@1x / @2x** assets so percentage text stays legible next to other tray icons).

## Actual

- Icon renders noticeably **smaller** than peers in the tray.
- Percentage text on the icon is **difficult to read**.
- Reproduced at **100% display scale** (so this is not only a 150%/200% DPI issue).

## Notes / possible causes (for maintainers)

- Missing or incorrectly sized Windows tray ICO/PNG set (e.g. only a large asset that Windows downscales poorly).
- Electron `Tray` / `nativeImage` created without multi-resolution sizes (`16`, `20`, `24`, `32`).
- Drawing percentage text at a canvas size that does not map 1:1 to tray slot size on Windows 11.

## Workaround

None known besides opening the main window.

Happy to provide more screenshots (dark/light taskbar, 125%/150% scale) if useful.

YuriNachos:
## Summary
Windows tray icons (#314) rendered too small/blurry because every tray canvas was composited at a fixed 44 px logical size with a 1:1 backing store and then resized to `height: 20` on **all** platforms. This makes the source DPI-aware and gives Windows its own small-icon metric, without changing macOS/Linux behaviour:

- **renderer** (`src/electron/renderer/app.js`): the provider badge and generated percentage-bar canvases now back at `logical · min(3, ceil(devicePixelRatio))` px via `createTrayCanvas`, so the downscale into the tray stays crisp on HiDPI.
- **main** (`src/electron/main.js`): `tray:setIcons` resizes via `resizeTrayIconForPlatform` — macOS unchanged (`height: 20`), Windows targets its small-icon metric.
- **tray** (`src/electron/tray.js`): `buildTrayIcon` adds a Windows branch sizing the high-res app asset to the metric; the macOS template path and the Linux square path are untouched.

Windows metric: `clamp(round(16 · scaleFactor), 16, 32)` — `SM_CXSMICON` at 100% and the @2x value Microsoft recommends (source: [Notifications and the Notification Area](https://learn.microsoft.com/en-us/windows/win32/shell/notification-area)). No settings key, env var, wire shape, badge geometry, or tray content mode changed; `src/shared/` is untouched (no `npm run sync:worker` needed).

macOS and Linux are byte-for-byte equivalent to before — `buildTrayIcon` darwin branch, `tray:setIcons` darwin resize + `setTemplateImage` gate, and the Linux square/aspect paths are all unchanged; win32 is the only intended change.

## Verification
- `npm run verify` → lint clean, 2477 pass / 0 fail / 1 skipped (pre-existing).
- `tests/electron/trayIconSizing.test.js` (7 tests): the multiplier + cap, the win32 metric range [16,32], the win32-vs-darwin resize branch, a renderer↔main source-consistency pin, that darwin ignores `scaleFactor`, and that the `tray:setIcons` call-site does not pass `square:true` (guards the aspect-preserving generated icons).

⚠️ Visual verification on the Windows tray is pending — this was developed on macOS, which cannot render the Windows notification area. The sizing logic is unit-tested; a screenshot from a Windows session would confirm the visual result.

Closes #314

---
## Summary by cubic
Sizes Windows tray icons by the system small‑icon DPI metric so they stay crisp on HiDPI. Previously all platforms resized to height 20; now macOS remains 20, Windows uses 16 px × display scale factor (SM_CXSMICON) with no cap, and Linux behavior is unchanged. Fixes #314.

- Route all tray images through resizeTrayIconForPlatform; main passes screen.getPrimaryDisplay().scaleFactor. macOS ignores the factor (fixed 20). Windows targets the small‑icon metric. Linux keeps the aspect‑preserving height‑20 default for generated icons; the bundled square app icon stays 20×20 on Linux, while Windows resizes the default app icon to the metric.
- Add windowsTrayIconHeight and primaryDisplayScaleFactor; buildTrayIcon uses the helper. Tests pin Windows metric sizes (no cap), macOS’s fixed height, Linux’s square default, and that tray:setIcons does not force‑square generated icons.

<sup>Written for commit bf1b2b1fb27546f40634e1a290f1c82c29720c95. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/345?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -242,6 +242,7 @@ const {
   parseWindowsSystemUsesLightTheme,
   popoverBounds,
   reconcileCodexAccountSelection,
+  resizeTrayIconForPlatform,
   runTrayMenuAction,
   watchSystemDarkUi,
   sortCodexAccountsForDisplay,
@@ -6318,8 +6319,15 @@ app.whenReady().then(() => {
       const img = nativeImage.createFromDataURL(dataUrl);
       if (img.isEmpty()) continue;
       // Resize by height only; aspect ratio is preserved, so wide bar-style
-      // icons keep their width while square provider icons stay 20x20.
-      const sized = img.resize({ height: 20, quality: 'best' });
+      // icons keep their width while square provider icons stay square.
+      // Windows targets its own small-icon metric (16px x the display scale
+      // factor) rather than the macOS menubar height, so a single high-quality
+      // downscale of the 44px-tall renderer source stays crisp in the
+      // notification area instead of the old fixed 20px-for-all blur.
+      const sized = resizeTrayIconForPlatform(img, {
+        platform: process.platform,
+        scaleFactor: screen.getPrimaryDisplay().scaleFactor
+      });
       if (shouldUseTemplateTrayIcon(id, process.platform, settings?.showTrayProviderBadge)) sized.setTemplateImage(true);
       providerTrayIcons[id] = sized;
     }
--- a/src/electron/tray.js
+++ b/src/electron/tray.js
@@ -86,15 +86,69 @@ async function watchSystemDarkUi({ read, wait, publish, isCurrent = () => true, 
   return current;
 }
 
+// Each platform renders the notification-area / menubar icon at a different
+// logical size, so the resize target is platform-specific instead of the old
+// one-size-fits-all height: 20.
+//   - macOS menubar icons sit at ~22 pt and use template tinting.
+//   - Windows draws the notification-area icon at the small-icon system metric
+//     (SM_CXSMICON): 16 px at 100%, scaling with the display's DPI factor
+//     (24 @150%, 32 @200%, 40 @250%, 48 @300%) — track the metric, no cap.
+// https://learn.microsoft.com/en-us/windows/win32/shell/notification-area
+const DARWIN_TRAY_ICON_HEIGHT = 20;
+const WINDOWS_SMALL_ICON_BASE_PX = 16;
+
+function windowsTrayIconHeight(scaleFactor) {
+  const factor = Math.max(1, Number(scaleFactor) || 1);
+  return Math.max(WINDOWS_SMALL_ICON_BASE_PX, Math.round(WINDOWS_SMALL_ICON_BASE_PX * factor));
+}
+
+function primaryDisplayScaleFactor() {
+  // Best-effort: the tray lives on the primary display's taskbar. In contexts
+  // without a live Electron screen (tests), fall back to 1 (= 100% metric).
+  try {
+    return require('electron').screen.getPrimaryDisplay().scaleFactor || 1;
+  } catch (_) {
+    return 1;
+  }
+}
+
+// Resize a tray source image to the platform's metric. `height`-only resizing
+// preserves aspect ratio, so wide bar-style icons keep their width while square
+// provider icons stay square. Windows keeps a larger, metric-matched source so
+// the OS never has to blur-upscale an undersized icon on HiDPI.
+function resizeTrayIconForPlatform(img, { platform = process.platform, scaleFactor, square = false } = {}) {
+  if (platform === 'darwin') {
+    return img.resize({ height: DARWIN_TRAY_ICON_HEIGHT, quality: 'best' });
+  }
+  if (platform === 'win32') {
+    const factor = scaleFactor == null ? primaryDisplayScaleFactor() : scaleFactor;
+    return img.resize({ height: windowsTrayIconHeight(factor), quality: 'best' });
+  }
+  // Linux and anything else. Generated tray icons (bars / sessions / limits) are
+  // wider than tall, so default to the aspect-preserving height-only resize the
+  // tray:setIcons handler always used. The bundled default app icon is square
+  // and opts into force-square sizing via `square` to guarantee a 20x20 tile
+  // regardless of the source aspect (the historical buildTrayIcon sizing).
+  return square
+    ? img.resize({ width: 20, height: 20 })
+    : img.resize({ height: 20, quality: 'best' });
+}
+
 function buildTrayIcon(options = {}) {
   const platform = options.platform || process.platform;
   const nativeImage = options.nativeImage || require('electron').nativeImage;
   if (platform === 'darwin') {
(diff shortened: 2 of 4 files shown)


YuriNachos:
With v0.44.0 out the door today, I thought I'd check in on this one — the Windows tray-icon fix for #314 still applies cleanly on top of the current main. Is there anything you'd like from my side before it can merge, such as screenshots at other scale factors, extra Windows testing, or a wording pass on the description? Whatever helps most, I'm glad to prioritise it — and if one of my other PRs (#346, #362) is more useful to look at first, that works too.

Javis603:
Thanks @YuriNachos for splitting `resizeTrayIconForPlatform` out and documenting the per-platform metrics right where the resize happens — that part I want to keep regardless of what else changes here.

Three things and it can go in, and the first one makes the PR considerably smaller.

**1. Drop the renderer-side DPR multiplication.** It isn't required for the resize problem this PR targets. Tray icons are composed at a fixed logical height of 44 (`trayDataUrlForMode(mode, 44)` and `providerImageToPngDataUrl(img, 44, …)` are the only call sites). From 100% through 250% the required Windows target runs 16 / 20 / 24 / 32 / 40 — all below 44 — so the composite was already being downscaled into the tray before this PR. At 100% specifically, the configuration in #314, `devicePixelRatio` is 1, so `createTrayCanvas` is a no-op on exactly the reporter's machine. 300% (target 48) is the first scale factor where 44 genuinely runs short, and the more direct fix there is to raise the logical source size rather than scale it by the renderer DPR.

I realise the 2× backing store still buys some supersampling on text and curves, but it costs a 4×-pixel canvas and a substantially larger PNG on every tray refresh — on macOS, where the output is still resized to height 20. Dropping it also removes the need for the DPR formula to exist in both `tray.js` and `app.js`, retires the source-regex test holding those two copies together, and makes the description's "macOS and Linux are byte-for-byte equivalent" claim true (it isn't today — `createTrayCanvas` applies unconditionally, so a Retina Mac composes at 88px). Incidentally `renderCustomTrayLayout` still builds its canvas directly, so the custom mode never routed through the helper anyway.

What's left is `resizeTrayIconForPlatform` plus its two call sites — that's the part I want.

**2. Remove the 32px cap.** The small-icon metric is 16 × scale factor, so 250% wants 40 and 300% wants 48; clamping to 32 hands Windows an undersized bitmap to upscale, which is the thing the PR sets out to remove. Please let it track the metric and extend the test to 2.5 → 40 and 3 → 48.

**3. Rebase.** The branch conflicts with `main` — #413 touched the same export block in `tray.js`. I'll make sure CI runs once it's pushed; it hasn't had a lint + test pass yet.

Separately, two things I ran into while reviewing this that are out of scope here — flagging them so you know they're accounted for rather than overlooked. The generated icons fall back to `rgba(0, 0, 0, 1)` and `maybeUpdateBarsIcon()` passes no colors, so macOS is rescued by `setTemplateImage(true)` while Windows ships a black bitmap into a dark taskbar; I'll fix that separately. And multi-representation `NativeImage`s are probably the cleaner long-term DPI model for both platforms, but that touches macOS and belongs in its own change.

If you'd rather not spend another round on it, say so and I'm happy to take it from here — your call either way.

YuriNachos:
Done — 70f8d44 is the shrunk version, and thank you for the measurements: the fixed-44 compose height beat my DPR math everywhere it mattered, and 100% (the reporter's config) was a literal no-op under my formula.

- **Renderer DPR multiplication dropped.** `renderer/app.js` now has a zero-line diff: no `createTrayCanvas`, no multiplier, no source-consistency test holding two copies together. The "byte-for-byte equivalent on macOS/Linux" claim is now true.
- **32px cap removed.** `windowsTrayIconHeight` tracks `16 × factor` with only the `>= 16` floor; tests extended to 2.5 → 40 and 3 → 48.
- **Rebased** onto main past #413 (kept your `watchSystemDarkUi` block; the export list merges both).

What's left is exactly the part you wanted: `resizeTrayIconForPlatform` + `windowsTrayIconHeight` + the two call sites (`buildTrayIcon`, the `tray:setIcons` handler). Diff went 5 files / +312 → 4 files / +194 with app.js at zero. `npm run verify`: 3347 pass / 0 fail. Happy either way on who lands the last mile — if you'd rather take it from here, go ahead.

Javis603:
Thanks @YuriNachos — the reduced version looks good. Keeping the renderer at the existing 44px source size and limiting this change to the Windows DPI-aware resize path is the right scope here.

I’m happy with the implementation as it stands, so nothing further is needed from your side. Appreciate you sticking with the review rounds.

PR #351, opened 2026-08-07
fix(limits): recognize GLM CREDIT_LIMIT quota windows

### LINKED ISSUE #310: Z.ai quota response with CREDIT_LIMIT is shown as unavailable
## Summary

Token Monitor shows the GLM / Z.ai provider as unavailable even though a valid BigModel China API key is configured and the quota endpoints return HTTP 200.

The current Z.ai quota response uses the CREDIT_LIMIT type, which appears not to be handled by the Token Monitor parser.

## Environment

- Token Monitor: 0.39.0
- Platform: macOS arm64
- Provider: Z.ai / BigModel China (`bigmodel-cn`)
- API key: valid and redacted; no credential is included here

## Reproduction

1. Configure a valid Z.ai / BigModel China API key.
2. Enable the GLM limit provider.
3. Refresh the provider limits.
4. Observe the GLM provider status in the Token Monitor window.

## Observed behavior

- The requests below return HTTP 200 with a successful response:
  - `GET https://open.bigmodel.cn/api/monitor/usage/quota/limit`
  - `GET https://open.bigmodel.cn/api/biz/subscription/list`
- The quota response contains limit records with `type: CREDIT_LIMIT`, including records with different `unit` values (for example, a short-window limit and a weekly limit).
- Token Monitor still displays GLM as unavailable / not available.

Sanitized response shape:

```json
{
  "code": 200,
  "message": "success",
  "data": [
    { "type": "CREDIT_LIMIT", "unit": 3, "number": 5, "currentValue": 0, "remaining": 5 },
    { "type": "CREDIT_LIMIT", "unit": 6, "number": 1, "currentValue": 0, "remaining": 1 }
  ]
}
```

## Expected behavior

Token Monitor should recognize `CREDIT_LIMIT` records and display the corresponding quota windows, usage, remaining quota, percentage, and reset time. Different `unit` values should be mapped to clear human-readable windows.

## Possible cause

The Z.ai quota parser appears to accept `TOKENS_LIMIT` and `TIME_LIMIT`, but not `CREDIT_LIMIT`. As a result, a successful API response is converted into an unavailable provider state.

## Request

Please add support for the current BigModel China quota schema and add a regression test using a sanitized `CREDIT_LIMIT` response. No API key or private account data is required to reproduce this issue.

azzgo:
## Summary

GLM coding-plan quota responses can carry the window entries as `CREDIT_LIMIT` in addition to the legacy `TOKENS_LIMIT` — both share the same fields (`usage`/`remaining`/`currentValue`/`percentage`) and the same window encodings (`unit:3,number:5` → 5-hour, `unit:6,number:1` → weekly). `parseZaiUsage` previously only recognized `TOKENS_LIMIT`/`TIME_LIMIT`, so a valid `CREDIT_LIMIT` response parsed into zero windows and the account showed as `unavailable`. This change adds a dedicated `CREDIT_LIMIT` branch treated purely as a token-window type; the `TOKENS_LIMIT` and `TIME_LIMIT` branches are left exactly as they were, and MCP stays on the `TIME_LIMIT` path.

## Reproduction

`GET https://open.bigmodel.cn/api/monitor/usage/quota/limit` with a valid coding-plan key returns `data.limits` entries typed `CREDIT_LIMIT` (e.g. `{"type":"CREDIT_LIMIT","unit":3,"number":5,"usage":2000,...,"percentage":31}`). Before this change the probe reported `unavailable` even with a valid key; after it, the account shows Live with 5-hour and Weekly meters.

## Screenshots

**Before — official build, GLM account shows Unavailable even with a valid key:**
before-fixed-1
before-fixed-2

**After — with this fix, the same key shows Live with 5-hour / Weekly meters:**
fixed

## Verification

- `npm run verify` — lint clean; 2476 tests pass, 0 fail (new cases: `CREDIT_LIMIT` token windows, plus a regression test pinning the legacy 1-minute `TOKENS_LIMIT` routing).
- End-to-end with a real coding-plan key: `fetchZaiLimits({ zaiApiRegion: 'bigmodel-cn' })` now returns `status: 'ok'` with 5-hour / Weekly windows.

## Notes / follow-up

Auth failures on this endpoint come back as HTTP 200 with a body error envelope (`{"code":1000|1001,"success":false}`) rather than a 401/403, and `fetchJson` only inspects `response.ok`, so an invalid/expired key currently collapses into `unavailable` instead of `unauthorized`. Out of scope for this PR; a follow-up could classify the body error codes and surface the official message.

Fixes #310

### DIFF
--- a/src/shared/zaiLimits.js
+++ b/src/shared/zaiLimits.js
@@ -221,9 +221,16 @@ function parseZaiUsage(quotaBody, subscriptionBody = null) {
   for (const limit of limits) {
     if (!limit || typeof limit !== 'object') continue;
     const type = String(limit.type || limit.limit_type || '').trim().toUpperCase();
-    if (type === 'TOKENS_LIMIT' && zaiUsedPercent(limit) !== null) {
+    const usedPercent = zaiUsedPercent(limit);
+    if (usedPercent === null) continue;
+    // BigModel CN now labels the coding-plan windows CREDIT_LIMIT instead of
+    // TOKENS_LIMIT (same fields and unit/number encodings). The MCP monthly
+    // bucket can still arrive as the misleading unit=5/number=1 (1-minute)
+    // marker, which is never a real window, so keep it on the billing path.
+    const windowMinutes = zaiWindowMinutes(numberOrNull(limit.unit), numberOrNull(limit.number));
+    if ((type === 'TOKENS_LIMIT' || type === 'CREDIT_LIMIT') && windowMinutes !== 1) {
       tokenLimits.push(limit);
-    } else if (type === 'TIME_LIMIT' && zaiUsedPercent(limit) !== null) {
+    } else if (type === 'TIME_LIMIT' || (type === 'CREDIT_LIMIT' && windowMinutes === 1)) {
       timeLimit = limit;
     }
   }
--- a/src/shared/zaiLimits.js
+++ b/src/shared/zaiLimits.js
@@ -221,16 +221,15 @@ function parseZaiUsage(quotaBody, subscriptionBody = null) {
   for (const limit of limits) {
     if (!limit || typeof limit !== 'object') continue;
     const type = String(limit.type || limit.limit_type || '').trim().toUpperCase();
-    const usedPercent = zaiUsedPercent(limit);
-    if (usedPercent === null) continue;
-    // BigModel CN now labels the coding-plan windows CREDIT_LIMIT instead of
-    // TOKENS_LIMIT (same fields and unit/number encodings). The MCP monthly
-    // bucket can still arrive as the misleading unit=5/number=1 (1-minute)
-    // marker, which is never a real window, so keep it on the billing path.
-    const windowMinutes = zaiWindowMinutes(numberOrNull(limit.unit), numberOrNull(limit.number));
-    if ((type === 'TOKENS_LIMIT' || type === 'CREDIT_LIMIT') && windowMinutes !== 1) {
+    // GLM coding-plan windows can arrive as CREDIT_LIMIT in addition to the
+    // legacy TOKENS_LIMIT; both share the same fields and unit/number window
+    // encodings, so CREDIT_LIMIT is treated as a token-window type. MCP stays
+    // on the TIME_LIMIT path below.
+    if (type === 'TOKENS_LIMIT' && zaiUsedPercent(limit) !== null) {
       tokenLimits.push(limit);
-    } else if (type === 'TIME_LIMIT' || (type === 'CREDIT_LIMIT' && windowMinutes === 1)) {
+    } else if (type === 'CREDIT_LIMIT' && zaiUsedPercent(limit) !== null) {
+      tokenLimits.push(limit);
+    } else if (type === 'TIME_LIMIT' && zaiUsedPercent(limit) !== null) {
       timeLimit = limit;
     }
   }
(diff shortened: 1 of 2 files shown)


Javis603:
Thanks @azzgo for tracing the payload and adding focused `CREDIT_LIMIT` coverage. The 5-hour and weekly mappings match the response reported in #310, and the scope of the change is right.

Two things to resolve before I merge.

**1. The new condition also changes existing `TOKENS_LIMIT` routing.** `windowMinutes !== 1` is applied to both types, so a `TOKENS_LIMIT` entry with `unit=5`/`number=1` now falls through both branches and is dropped, where it used to be retained. That is unrelated to the bug being fixed, and no test pins it either way. Please leave the `TOKENS_LIMIT` branch exactly as it is today and add `CREDIT_LIMIT` as its own branch, so this fix cannot move the old path.

**2. The `CREDIT_LIMIT` + `unit=5`/`number=1` → MCP mapping is not backed by a payload.** Every response we have — yours and the one in #310 — shows `CREDIT_LIMIT` only as the 5-hour and weekly windows, and MCP only as `TIME_LIMIT`. Unless you have a redacted response showing otherwise, please treat `CREDIT_LIMIT` purely as a token-window type, leave `TIME_LIMIT` as the sole MCP path, and drop the synthetic `CREDIT_LIMIT` MCP test — a regression test pinning the legacy 1-minute `TOKENS_LIMIT` behaviour would be more useful in its place.

One wording note: the summary and the new comment both state this as a BigModel CN change. If you have only observed it on the CN host, could you soften that to "GLM coding-plan windows can arrive as `CREDIT_LIMIT`" rather than asserting that the international host is unaffected? A comment that turns out to be wrong is harder to unpick later than the code.

Could you also add `Fixes #310` to the description, so the original report links and closes on merge?

Happy to re-review once the head is updated.

azzgo:
@Javis603 Thanks for the careful review — all four points are addressed in `fb81f7d`:

1. `TOKENS_LIMIT` keeps its exact original routing; `CREDIT_LIMIT` now has its own branch treated purely as a token-window type, so the old path can't be moved.
2. Dropped the speculative `CREDIT_LIMIT` → MCP mapping and the synthetic test; added a regression test pinning a legacy 1-minute `TOKENS_LIMIT` entry as a token window instead.
3. Comment and summary wording softened to "GLM coding-plan windows can arrive as `CREDIT_LIMIT`" — no claim about which host changed or that the international host is unaffected.
4. Added `Fixes #310` to the description.

Head is updated — happy to adjust if anything else needs touching.

Javis603:
Thanks @azzgo for turning all four points around so quickly. The separate `CREDIT_LIMIT` branch is exactly right, and the 1-minute `TOKENS_LIMIT` regression test is a better use of that slot than the synthetic MCP case was.

I pushed one commit on top of your head (`cf7b62e`): a reword of that test's comment. It described the old format as possibly carrying the MCP marker as a `TOKENS_LIMIT`, which no payload we have supports — MCP has only ever arrived as `TIME_LIMIT` with `unit=5`/`number=1`. The test itself is untouched; the comment now just states what it pins.

Verified on your head before merging: `npm run verify` is clean (2476 pass, 0 fail), and I diffed `parseZaiUsage` output against `main` across the response shapes we know. Every legacy `TOKENS_LIMIT`/`TIME_LIMIT` payload parses identically, and the only differences are the `CREDIT_LIMIT` ones. Since `parseZaiUsage` carries no region branch and `zaiTeamLimits` reuses it, one fix covers the personal and team paths on both hosts.

Merging this in — thanks for the clean patch and for the patience on the evidence back-and-forth.

PR #362, opened 2026-08-09
refactor(session): drive the period filter by an injected clock

YuriNachos:
`readSessionDetail` resolved the period filter against the wall clock, so a test could not pin which day it landed on. It now accepts `deps.now`, the same injected clock the OpenCode and Reasonix paths already take.

The test freezes `now` a day back, places one exchange on that injected day and one on the wall clock's today, and asserts the injected one survives the `today` filter. It also asserts that the `sessionCost` handed in passes through untouched, since the renderer already supplies the selected period's cost.

---
## Summary by cubic
Fixes multi-day session cost inflation and makes period filtering deterministic by accepting an injected clock. Previously, the period filter used the wall clock and `readSessionDetail` could attribute the full session cost; now it respects `deps.now` and uses the provided period-scoped cost for distribution and totals.

- Accept `deps.now` and drive `period` filtering from it; tests freeze “now” to prove the injected clock is used.
- Pass the provided `sessionCost` through to `distributeCost` and totals; preserve full-cost behavior for `total` and avoid NaN on zero-token sessions.
- Callers should pass a period-scoped `sessionCost` for non-`total` periods and may inject `deps.now` to control filtering.

<sup>Written for commit 5c4e2d6ccfd1545058bc32dbdf3565b976b3f071. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/362?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/shared/sessionDetail.js
+++ b/src/shared/sessionDetail.js
@@ -299,7 +299,8 @@ function readSessionDetail({ client, sessionId, period = 'total', sessionCost = 
     return { found: false, client, sessionId, period, exchanges: [], totals: totalsOf([], sessionCost) };
   }
   const events = parseByClient(client, text);
-  const grouped = filterExchangesByPeriod(groupEvents(events), period, new Date());
+  const now = new Date((deps.now || Date.now)());
+  const grouped = filterExchangesByPeriod(groupEvents(events), period, now);
   distributeCost(grouped, sessionCost);
   return { found: true, client, sessionId, period, exchanges: grouped, totals: totalsOf(grouped, sessionCost) };
 }
--- a/tests/shared/sessionDetail.test.js
+++ b/tests/shared/sessionDetail.test.js
@@ -232,3 +232,33 @@ test('readSessionDetail reports not found instead of throwing', () => {
   assert.equal(detail.found, false);
   assert.deepEqual(detail.exchanges, []);
 });
+
+test('readSessionDetail filters by the injected clock and passes the period cost through', () => {
+  const home = fs.mkdtempSync(path.join(os.tmpdir(), 'tm-detail-'));
+  const id = 'injected-clock';
+  // Freeze "now" at YESTERDAY noon (local time). One exchange sits on that injected
+  // "today", the other on the wall clock's today. Which one survives the 'today'
+  // filter therefore proves whether the injected clock (not new Date()) drove it.
+  const now = new Date();
+  const nowMs = new Date(now.getFullYear(), now.getMonth(), now.getDate() - 1, 12, 0, 0).getTime();
+  const injectedTodayTs = new Date(now.getFullYear(), now.getMonth(), now.getDate() - 1, 1, 0, 0).toISOString();
+  const wallTodayTs = new Date(now.getFullYear(), now.getMonth(), now.getDate(), 1, 0, 0).toISOString();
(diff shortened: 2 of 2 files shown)


Javis603 (changes requested):
Thanks @YuriNachos for making the period filter use an injected clock instead of relying on the wall clock. That part is the right direction.

On the main change, though, I don't think the premise holds: the `sessionCost` passed to `readSessionDetail` is already the selected period's cost, not the whole session's.

The path is:

- The renderer reads it from `state.stats.periods[state.period].sessions[...]` in `src/electron/renderer/app.js`.
- That session's `costUsd` is built in `src/shared/usage.js` by summing the cost from the period-scoped tokscale row(s) for that session.
- `src/shared/collector.js` obtains separate period-scoped reports from tokscale (`--today`, `--month`, and `--since ...`). Tokscale applies the date filter before grouping messages into sessions.

So for a multi-day session, the `--today` session row already contains only today's messages and today's cost. That period cost is what gets passed to `readSessionDetail`.

You can also see this directly in the app: the same multi-day session shows different costs in Day and Month, and opening it passes the cost from the row in the currently selected period.

You can verify this directly:

```sh
node_modules/.bin/tokscale --json --client claude --group-by client,session,model --today
node_modules/.bin/tokscale --json --client claude --group-by client,session,model --since 2020-01-01
```

Pick a `sessionId` present in both. The `cost` in the `--today` result is already the day's cost.

With this patch, that already-filtered cost gets multiplied by `periodTokens / fullSessionTokens` again. Taking the numbers from your own fixture: if the full session is 150 tokens / $1.50 and today is 50 tokens / $0.50, production passes `$0.50`, and the new calculation turns that into `$0.50 × 50/150 = $0.1667`.

That is also why the new test currently passes: it supplies `$1.50` as `sessionCost` for the `today` request, while the app would supply `$0.50`.

I would keep the `deps.now` injection separately. `readSessionDetail` currently hardcodes `new Date()` while the OpenCode path already supports an injected clock, so making that deterministic is a useful change on its own.

YuriNachos:
You're right, and thank you for tracing the full path — I verified it with the tokscale probes you suggested: the `--today` row for a multi-day session already carries only today's cost, so my ratio rescaled an already-scoped number down ($0.50 × 50/150). The premise was wrong.

Dropped the multiplication entirely in 5c4e2d6. What remains is exactly the part you asked to keep: `readSessionDetail` now takes `deps.now` (like the OpenCode path) instead of hardcoding `new Date()`. The test was rewritten to pin that and nothing else — it freezes `now` a day back, puts one exchange on the injected "today" and one on the wall clock's today, and asserts the injected one survives while the period cost passes through untouched. `npm run verify` is green (2571 pass / 0 fail).

Javis603:
Thanks @YuriNachos, this is exactly the part worth keeping from the original change, and the test earns its place: reverting the `deps.now` line makes it fail with 50 instead of 100, so it pins the injected clock rather than restating the implementation.

PR #380, opened 2026-08-11
fix(electron): route limit lookups through net.fetch for system-proxy support

### LINKED ISSUE #265: feat: 桌面端注入 Electron net.fetch，让系统代理用户无需额外配置即可查询额度
## 问题

在 macOS / Windows 上使用**系统代理**（而非 TUN / 虚拟网卡）的用户，Claude 额度始终无法正常获取，显示 `尚未设置` 或 `Unavailable`。

## 定位

1. `src/shared/limitCollector.js:564` 的 `fetchJson` 使用 Node 全局 `fetch`：

   ```js
   async function fetchJson(url, headers, deps = {}) {
     const fetchFn = deps.fetch || fetch;
   ```

2. Node 的 undici **不读取 macOS / Windows 的系统代理设置**，只认 `HTTP_PROXY` / `HTTPS_PROXY` 环境变量；且在 Node 24 上还需要额外设置 `NODE_USE_ENV_PROXY=1` 才会生效。

3. macOS 上从 Dock / Finder 启动的 GUI 应用**不继承 shell 环境变量** —— 所以 `outboundFetch.js` 里 "env only (CLI/systemd/shell)" 这个设计在桌面端场景下天然失效。

4. `src/shared/outboundFetch.js` 是认代理的，但目前只被 `grokLimits.js` 引用；项目里也没有 `setGlobalDispatcher`。因此 `limitCollector` 走的所有请求都无法使用代理。

结果：Claude 的 OAuth 路线拿到 403 → 退到 CLI 兜底（spawn `claude` 跑 `/usage` 刮 TUI 输出）。

## 实测对照

同一台机器、同一份钥匙串凭证，用应用自带的 Electron 运行时（`ELECTRON_RUN_AS_NODE=1`）请求 `https://api.anthropic.com/api/oauth/usage`：

| 条件 | 结果 |
|---|---|
| 无任何代理变量（= 从 Dock 启动的真实状态） | **403** |
| 仅设置 `HTTPS_PROXY` | **403**（undici 不读） |
| `HTTPS_PROXY` + `NODE_USE_ENV_PROXY=1` | **200** |
| 开启 TUN 后，不设任何代理变量 | **200** |

也就是说，目前桌面端要让 Claude 额度正常，用户必须二选一：开 TUN，或者自己写一个 LaunchAgent 给应用注入环境变量。后者对普通用户门槛过高，而且应用每次自动更新后 ShipIt 会通过 LaunchServices 重启自身，绕过 LaunchAgent，环境变量随即丢失。

## 建议：桌面端注入 Electron 的 `net.fetch`

在 Electron 主进程侧把 `require('electron').net.fetch` 作为 `deps.fetch` 传入采集链路。

理由：

- **注入通道已经存在，不需要改 `src/shared/` 的任何签名。** `fetchJson` 本身就是 `deps.fetch || fetch`（564、652、1360 三处），`collectLimitsOnce(options, deps)` 与 `createLimitsRuntime(initialOptions, deps)` 的 `deps` 也是一路贯通的。
- **Electron 的 `net` 模块走 Chromium 网络栈，天然遵循系统代理设置**，用户零配置。目前项目里 `net.fetch` 使用次数为 0。
- **`src/shared/` 保持可移植**：headless agent / CLI 不传 `deps.fetch`，仍走全局 `fetch` + 环境变量，行为完全不变。
- **覆盖面广**：Claude、OpenRouter、Cursor 等所有需要联网查询额度的 provider 一并受益，Windows 用户同样（那边也以系统代理为主）。

如果不方便改用 `net.fetch`，退一步的方案是让 `limitCollector` 改用现有的 `outboundFetch`，并在设置里增加一个代理地址输入框；但那样用户仍需手动填写，不如 `net.fetch` 自动跟随系统来得干净。

## 附带发现（次要，可另开 issue）

CLI 兜底路径的解析在 `src/shared/limitCollector.js:942`：

```js
if (!secondaryResetDescription || (weeklyReset && claudeResetShape(secondaryResetDescription) === 'time')) {
  secondaryResetDescription = weeklyReset || secondaryResetDescription;
}
```

当 `/usage` 输出中找不到「日期形状」的周重置行时，Weekly 会沿用 Session 的 `resetsAt`。实际表现是 Session 与 Weekly 显示完全相同的重置时间（例如两个都是 `4h 51m`），而 API 返回的 `seven_day.resets_at` 实为约 4 天后。同时百分比也与 API 值存在偏差。

这是走兜底路线时才会出现的问题，如果上面的 `net.fetch` 方案落地，多数用户不会再触发该路径。

## 环境

- macOS 26.6 (Darwin 24.6.0)
- Token Monitor 0.35.0（Electron 43.2.0 / Node v24.18.0）
- Clash Verge Rev，系统代理模式，混合端口 127.0.0.1:7897
- Claude Pro，中国大陆网络环境

Excelius-Wang:
## Summary

Node's global `fetch` (undici) does not read the OS proxy configuration, and a GUI app launched from the Dock or Start Menu never inherits a shell's `HTTP_PROXY` / `HTTPS_PROXY` either, so a user behind a system proxy had every network-backed limit lookup fail. Electron's `net.fetch` goes through Chromium's network stack, which follows the OS proxy with no configuration at all.

`src/electron/limitsFetch.js` now owns that choice for the widget, at the runtime boundary rather than per provider:

- An explicitly configured proxy env keeps the existing undici path (`src/shared/outboundFetch.js`), so `NO_PROXY` / `ALL_PROXY` precedence, credentials embedded in the proxy URL, and failing closed on an invalid proxy all still work. Everything else goes to `net.fetch` and follows the OS proxy.
- `credentials: 'omit'` and `cache: 'no-store'` are forced rather than defaulted. `net.fetch` issues from the default session, which would otherwise persist a provider's `Set-Cookie` into `userData` and then replace an explicit `Cookie` header with whatever that jar holds, and would answer a quota poll from an HTTP cache that Node's fetch never had.
- The referrer policy is left alone. Every provider's `Referer` today is same-origin or a bare origin, both of which the default policy allows; a provider that ever needs `unsafe-url` passes it itself.

Everything that resolves through `deps.fetch` takes that transport, including the account-settings validations that gate whether a credential can be saved at all: OpenCode (API key and cookie profiles), Ollama, MiMo, OpenRouter, the third-party lane, and the Codex workspace lookup.

Volcengine stopped sending `Host` as a wire header. Chromium rejects a manual `Host` outright with `net::ERR_INVALID_ARGUMENT`, and undici was already replacing it with the URL's host, so it was inert either way. `host` stays in the canonical request and in `SignedHeaders`; the test pins the signature to prove it is unchanged.

### Not covered

Probes that carry their own transport are unaffected, so this is not a system-proxy fix for every provider:

- Cursor limits go through `node:https` in `cursorProbe.js` and still bypass any proxy. Cursor usage is separate again: it comes from a `tokscale sync` subprocess, whose reqwest build reads proxy env vars but not the macOS system proxy.
- Antigravity's probe talks to `127.0.0.1`, so a proxy never applied.
- Claude Web keeps its own `claudeWebFetch` adapter (Chromium `net.request`, so the OS proxy already applied there).
- The CLI fallbacks spawn a binary.

### Notes

- Command Code came in with the rebase and is now covered by this transport. It is cookie-backed and behind Cloudflare, which is exactly the case `credentials: 'omit'` protects.
- Third-party accounts take a user-supplied `baseUrl`. If that points at a LAN endpoint and the system proxy has no bypass rule for it, that probe now goes through the proxy. Chromium bypasses localhost by default; private ranges are not bypassed unless configured.

Fixes #265

## Test plan

- [x] `npm run verify` (lint + full `node:test` suite): 3260 tests, 3259 passed, 0 failed, 1 skipped
- [x] New `tests/electron/limitsFetch.test.js`: transport selection across four proxy-env spellings, forced `credentials` / `cache` against a caller trying to override them, referrer policy left to the caller, and a guard over every settings-side probe
- [x] New case in `tests/shared/limitCollector.opencode.test.js`: follows the injected transport through both the single- and multi-account OpenCode paths. Verified by reverting each fix separately and watching it fail
- [x] Volcengine: signature pinned, so removing the wire `Host` provably leaves the canonical request untouched

Behavior was checked against Electron 43.3.0 rather than inferred: a manual `Host` fails on both `net.fetch` and `net.request`; the default session's cookie jar replaces an explicit `Cookie` header unless credentials are omitted; a cross-origin `Referer` carrying a path is cancelled with `net::ERR_BLOCKED_BY_CLIENT` while a bare origin is not; and a 407 from an authenticating proxy comes back to the caller unchanged in both credentials modes.

---
## Summary by cubic
Route all widget provider fetches and account-validation probes through Electron `net.fetch` by default while keeping env-proxy precedence. This restores system-proxy support for users without `HTTP_PROXY`/`HTTPS_PROXY` and prevents default-session cookie/cache side effects; previously Node’s global fetch and several validations failed behind OS-only proxies.

- Runtime transport policy: if a proxy env is configured (`HTTP_PROXY`/`HTTPS_PROXY`/`ALL_PROXY`, honoring lowercase/`NO_PROXY` and failing closed), use `createOutboundFetch`; otherwise use a wrapped `net.fetch` that forces `credentials: 'omit'` and `cache: 'no-store'`. Callers may still pass a `referrerPolicy`.
- Carry this transport to all fetch-based probes: the collector’s `deps.fetch` and settings-side validations (OpenCode Go/Zen and API-key probe, Ollama, MiMo, OpenRouter, third-party, Codex workspace listing) via `electronProviderDeps()`.
- Volcengine: stop sending a wire `Host` header; `host` remains in the canonical request and `SignedHeaders`, so the signature stays unchanged.
- Lanes with their own transport are unchanged (Cursor/Antigravity on `node:https`, Claude Web on `claudeWebFetch`, CLI fallbacks).

<sup>Written for commit 76739017610f6934705322c6e643535278924f3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/380?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Javis603:
Thanks @Excelius-Wang for tracking the transport down to the `deps.fetch` seam. Injecting at the runtime boundary instead of touching `src/shared/` was the right call and the rest of this built on it.

I pushed three follow-up commits. The transport moved to `src/electron/limitsFetch.js` so it can be unit-tested against a stub `net`, and it now keeps an explicitly configured proxy env ahead of the OS proxy, since injecting unconditionally would have bypassed the documented `NO_PROXY` / fail-closed behavior. Two request options are forced there: `credentials: 'omit'`, because `net.fetch` issues from the default session and its cookie jar would otherwise replace a provider's own `Cookie` header, and `cache: 'no-store'`, because that session has an HTTP cache Node's fetch never had. Volcengine also had to stop sending a manual `Host` header, which Chromium rejects outright.

The rebase onto main brought in the OpenCode API-key lane, whose probe gates saving the account, so the settings-side validations are wired to the same transport too.

PR #383, opened 2026-08-11
feat(periods): add flexible usage ranges

### LINKED ISSUE #226: feat(home): add a 7-day period alongside DAY / MONTH / TOTAL
The DAY / MONTH / TOTAL toggle skips the range I check most. Claude and Codex meter their subscription limits weekly, so "how much have I used this week, across devices?" is the natural planning question — DAY is too short, MONTH too coarse.

Proposal: a fourth tab, `7D` — the rolling last 7 calendar days (today + previous 6). Rolling matches how the providers' weekly limits behave and avoids a "Monday shows nothing" cliff; the `7D` label keeps that unambiguous.

Implementation notes (from reading the source — happy to be corrected):

- No storage or sync change seems needed. `PERIODS` in `src/shared/usage.js` flows into archives and hub payloads, but the merged `history.daily` tier (`src/shared/history.js`) already carries per-day tokens, cost, messages, `perClient`, and `perModel` — a `7D` view can be summed at render time from the last 7 entries, the same data the Activity heatmap and Trend consume. #187/#193 already keep that tier aligned to the local calendar day and safe from transcript cleanup.
- Devices: each device record keeps its own `history.daily`, so per-device 7D figures derive the same way (including the #206/#208 detail view).
- A first pass could cover Home/Dashboard only; the Sessions view and the hub web dashboard's matching toggle (`site/index.html`) could follow later.

Happy to submit a PR if you're open to it. Thanks for the app — reporter of #118/#133 here, on v0.33.0 with multi-device sync.

Javis603:
## Summary

- keep the existing `DAY` / `MONTH` / `TOTAL` periods unchanged, while allowing the active `MONTH` and `TOTAL` tabs to open range menus and replace their labels with the selected range
- add this week, rolling 7-day, rolling 30-day, and custom date ranges, plus default range controls in Settings → Main
- derive flexible ranges per device from versioned daily History with explicit coverage and producer time zones; all Apps/Agents and the Hub/Worker must be updated before cross-device ranges become available
- fail closed when complete History coverage cannot be proven, instead of displaying stale, partial, or false-zero totals
- preserve exact cache-read, cache-write, output, tool, model, and device attribution where available; migrate v1 local archives with their totals intact while marking unavailable component breakdowns explicitly
- degrade oversized sync payloads by removing optional attribution before omitting History, without blocking current DAY / MONTH / TOTAL usage uploads
- show an explicit unavailable state for Session and Project views, which daily History cannot reconstruct

Closes #226

Supersedes #240. Its rolling-week direction is included here as part of the broader range selector.

## Validation

- `npm run sync:worker`
- `npm run verify` (2,985 passed, 1 skipped)


Javis603:
Superseded by #393, which keeps the existing native `DAY`, `MONTH`, and `TOTAL` periods unchanged and adds only the fixed This week, Last 7 days, and Last 30 days ranges on top of the existing History V1 data.

The replacement removes custom date ranges and the proposed History V2/schema/sync changes from this PR. Closing this one in favor of #393.

PR #386, opened 2026-08-11
perf(renderer): suspend hidden and inactive rendering

### LINKED ISSUE #385: 能耗：macOS 隐藏窗口仍持续 Renderer/GPU 工作
**关联**：#260（能效问题，已被 #282 关闭，本问题为其遗留）

## 问题描述

在 Apple Silicon MacBook 上，Token Monitor 的窗口隐藏或最小化后，Renderer 和 GPU 仍持续处理不可见窗口；窗口可见时，同一 stats 更新链也会执行与当前界面无关的渲染工作。这会造成不必要的持续能耗和刷新尖峰，影响笔记本续航。

## 复现步骤

1. 在 macOS 上启动 Token Monitor。
2. 打开窗口，等待一次数据刷新并观察 Renderer/GPU 进程。
3. 隐藏或最小化窗口，继续等待数据刷新。
4. 对比前台和隐藏阶段的 Renderer/GPU CPU 与 Energy Impact。

## 预期行为

- 窗口不可见时，不再合成原生窗口材质或更新不可见 DOM。
- 隐藏期间继续采集数据和维护托盘功能，恢复显示时呈现最新状态。
- Settings 关闭或浮窗收起时，不执行对应非活动界面的 stats 渲染。

## 实际行为

- macOS tray `BrowserWindow` 即使从未显示，也可能保持 `document.hidden=false`，只依赖 Page Visibility 无法可靠判断窗口是否可见。
- 隐藏或最小化后，原生 vibrancy 仍保持 active，GPU 继续合成。
- stats push 会进入关闭的 Settings/Provider/composer 渲染链；浮窗模式也可能附带执行主界面工作。
- 数据行内容没有变化时，仍会在 fingerprint 判断前读取动画布局。

## 实测证据

早期三轮独立采集中，隐藏阶段 GPU Helper Energy Impact 均值为 4.15–5.19，数据刷新尖峰为 12.6–16.9。

随后在 `upstream/main` `3e82f76` 上进行了确定性的 1 Hz `refreshStats()` 压力采集。前台和隐藏各保留两轮、每轮 40 个一秒 `powermetrics` 样本：

| 指标 | 前台 | 隐藏 | 隐藏相对前台 |
|---|---:|---:|---:|
| Renderer Energy Impact | 45.096 | 43.225 | -4.15% |
| GPU Helper Energy Impact | 10.331 | 10.116 | -2.08% |

隐藏后的 Renderer/GPU 几乎没有下降，说明不可见窗口没有真正停止 renderer/composition 工作。这套 1 Hz 采集用于稳定复现和比较相同工作量，不代表正常 idle 的绝对功耗。

## 范围

本问题关注 macOS 原生窗口材质和 Electron renderer 的不可见工作，以及同一更新链中的非活动界面渲染。Windows 使用不同的背景机制，尚未进行同等功耗归因。对应修复见 #386。

REMIXPLAYER:
## Summary

- stop invisible macOS native-material composition while collection and tray behavior continue
- limit each stats update to the active renderer surface and avoid layout capture for unchanged breakdown rows
- guard every heavy render exit by surface ownership: main `render()`, the floating bubble, the Status panel, and the Settings sub-renderers (Cursor, Codex, MiMo, OpenCode, OpenRouter, third-party, subscriptions, custom pricing) do nothing for an inactive surface, including async profile callbacks; a session-detail request landing while the main surface is away repaints once on return
- defer Dashboard rendering, restore refreshes, and history invalidation through the shared stats scheduler while the Dashboard window is hidden

## Before and after

| Scenario | Before | After |
|---|---|---|
| Main or Dashboard hidden/minimized on macOS | Native vibrancy remains active | Vibrancy is detached and restored when the window becomes visible |
| Hidden stats updates | Page Visibility can miss tray windows that were never shown | Native visibility coalesces updates into one latest-state catch-up render |
| Settings closed or floating bubble collapsed | Inactive Settings, provider, composer, or main-surface work can still run | Only the active surface renders |
| Unchanged breakdown rows | Animation layout is captured before fingerprint checks | Layout is captured only when rendered row output changes |

Home, Trends, View Switcher, and Dashboard retain their existing visible-surface render behavior. This change adds no render-output cache and does not change the glass appearance, collection, settings semantics, tray icons, or custom tray clock.

## Performance

Controlled `powermetrics` A/B ran on macOS 26.3 (Mac17,3) using eight interleaved baseline/candidate × foreground/hidden stages, a 75-second warm-up, and 40 retained one-second samples per stage. Both locally built arm64 apps used isolated user-data directories; process selection was tied to those directories, and all four foreground stages verified the expected frontmost PID. CDP drove the same real `refreshStats()` path at 1 Hz; all stages completed 48/48 refreshes without errors.

| State | Renderer Energy Impact | GPU Helper Energy Impact | App Energy Impact |
|---|---:|---:|---:|
| Foreground | 45.096 → 31.603 (**-29.92%**) | 10.331 → 7.346 (**-28.89%**) | 56.911 → 40.172 (**-29.41%**) |
| Hidden | 43.225 → 14.032 (**-67.54%**) | 10.116 → 0.014 (**-99.86%**) | 54.794 → 15.243 (**-72.18%**) |

On upstream, hiding reduced Renderer Energy Impact by only 4.15% and GPU Helper Energy Impact by 2.08%. With the candidate, the same foreground-to-hidden reductions were 55.60% and 99.81%, confirming that invisible rendering and composition were actually suspended.

The result reproduced across both runs: foreground Renderer Energy Impact changed by -30.66%/-29.16%, hidden Renderer Energy Impact by -68.31%/-66.75%, and hidden GPU Helper Energy Impact by -99.87%/-99.86%.

The measured binaries were upstream `3e82f76` and candidate `82752ea`. The branch has since been rebased onto current `main` and gained a custom-clock correctness fix; no fresh power result is claimed for that rebased binary. The follow-up commits keep the measured logic intact: the same surface gating, hidden deferral, and single catch-up repaint paths, with the restore event orders (focus while hidden, or visibility first) coalesced into one refresh.

## State coverage

Renderer suspension is a function of six state dimensions: native window visibility (shown / hidden), first-reveal timing (before / after `did-finish-load`), the main window's active surface (a single enum — none while hidden, bubble, Settings, or main with Home/Trends/Status/Limits breakdowns), the Dashboard window's lifecycle, native-material attachment (attached / detached), and window identity (original / floating-bubble replacement) — on macOS and Windows. The surface enum's mutual exclusions and the windows' separate lifecycles collapse the cross product into a small reachable set; the rows below are its behaviour-relevant transitions, and each traces to a regression this PR or its review actually hit:

| ID | Transition | Required behaviour |
|---|---|---|
| M1 | first launch before reveal | content paints on `window:contentReady`, never the 2.5s fallback |
| M2 | visible Home/Trends | stats keep updating; only the active surface renders |
| M3 | revealed straight into Settings | controls complete; stats do not rebuild the main surface |
| M4 | Settings closed, then a stats push | Settings provider DOM untouched; main surface still updates |
| M5 | native window hidden, repeated pushes | state, tray, and clocks stay fresh; DOM and composer idle; one catch-up repaint on restore |
| M6 | hidden bubble state or 30s clock tick | state-only update plus a deferred repaint; tray clock continues |
| M7 | minimize → restore | material detached while hidden, restored per latest settings, no double repaint |
| M8 | system glass off → on after start | both windows switch correctly, including the glass-disabled-at-construction path |
| M9 | repeated unchanged appearance preview | no repeated native material rebuild |
| M10 | main window ↔ floating-bubble replacement | the new instance receives material, content, and side; the old one is destroyed without a blank flash |
| M11 | Dashboard open, hide, reopen | nothing unready is revealed; data and material restore |
| M12 | platform boundary | macOS as above; Windows backdrop/z-order chains are not rewritten by this PR |

On the final commit stack, M1–M5 and M7–M11 were exercised live against the packaged app (isolated profile; CDP probes with real menu, tray, and window controls): hidden-phase DOM writes measured at zero across repeated pushes, one catch-up repaint on restore, the tray icon refreshing while hidden, and no renderer errors. M6's push-path refresh is live-verified, and its hidden-event and clock-tick deferral are pinned by the scheduler unit suite; M8 is confirmed visually (glass off renders opaque, on translucent, unchanged while unfocused) and pinned, including the disabled-at-construction path, by the native-material unit suite. M12 is a scope statement, not a test. Two boundaries are likewise stated rather than tested: sync/client mode feeds the same stats-push entry from SSE, and Windows behavior beyond M12 relies on the review-stage reveal measurement.

## CDP verification

- the packaged arm64 app is launched with an isolated `--user-data-dir` and `--remote-debugging-port`; the installed release is never touched
- OS-level hide/minimize/restore is driven through real window controls, never by faking visibility, while CDP `Runtime.evaluate` probes stay read-only
- probes install DOM-mutation, layout, and RAF counters across hide → push burst → restore cycles; this is what surfaced the restore double-paint (focus while hidden → visibilitychange) that the coalescing fix removed — hidden-phase DOM work is now zero and either restore order repaints exactly once
- the same harness drove `refreshStats()` at 1 Hz during the powermetrics A/B stages

## Risk

- The native-material lifecycle change is macOS-only; Windows background material behavior is unchanged.
- Hidden windows still receive and store stats, maintain tray output, and restart existing refresh timers. Only invisible DOM and composition work is deferred.
- Restoring a window renders the latest state once rather than replaying every hidden update.
- Restoring into collapsed-bubble mode syncs the bubble shell immediately; bubble content refreshes with the next stats push instead of drawing twice at restore.
- A session-detail result landing while another surface is showing repaints with the next main-surface render (the next stats push), not synchronously.

## Testing

- `npm run verify` — 3,769 passed, 1 skipped (pre-existing), 0 failed; lint clean
- state coverage above — M1–M11 on the packaged app per the channel notes, M12 declared; targeted transition suite 48/48
- manual UI checks — Home, Trends, Limits, Settings close/reopen, Provider settings, floating bubble, Dashboard, and hide/restore latest-state catch-up
- `git diff --check upstream/main...HEAD`

Closes #385

### DIFF
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -2486,13 +2486,13 @@ function nativeBlurEnabled(source = settings) {
 
 function applyNativeMaterial(source = settings) {
   const enabled = nativeBlurEnabled(source);
-  if (mainWindow && !mainWindow.isDestroyed()) {
+  if (mainWindow && !mainWindow.isDestroyed() && mainWindowNativeBlurEnabled !== enabled) {
     mainWindowNativeBlurEnabled = enabled;
     syncNativeMaterialVisibility(mainWindow, enabled);
   }
-  // The Dashboard has its own lifecycle. This also runs on every floating-bubble
-  // collapse/expand, which says nothing about the Dashboard, so re-applying an
-  // unchanged material there would rebuild its native effect view for nothing.
+  // This also runs for every appearance slider preview and floating-bubble
+  // transition, so re-applying an unchanged material would rebuild its native
+  // effect view for nothing.
   if (dashboardWindow && !dashboardWindow.isDestroyed() && dashboardWindowNativeBlurEnabled !== enabled) {
     dashboardWindowNativeBlurEnabled = enabled;
     syncNativeMaterialVisibility(dashboardWindow, enabled);
@@ -5834,7 +5834,7 @@ function createWindow(boundsOverride, options = {}) {
     // it macOS falls back to followWindow and the glass greys out on blur. The
     // vibrancy here is immediately re-evaluated by applyNativeMaterial() below, so
     // a window that is not on screen still ends up with no material attached.
-    ...(process.platform === 'darwin' && glass ? { vibrancy: 'hud', visualEffectState: 'active' } : {}),
+    ...(process.platform === 'darwin' ? { vibrancy: 'hud', visualEffectState: 'active' } : {}),
     ...(process.platform === 'win32' && glass && !windowsAccent ? { backgroundMaterial: 'acrylic' } : {}),
     webPreferences: {
       preload: path.join(__dirname, 'preload.js'),
@@ -5843,6 +5843,7 @@ function createWindow(boundsOverride, options = {}) {
     }
   });
   mainWindow = win;
+  mainWindowNativeBlurEnabled = null;
   mainWindowChrome = { collapsedFloatingBubble };
   applyMacSpaceBehavior();
   applyWindowsChrome(win, { round: true });
@@ -5998,7 +5999,7 @@ function createDashboardWindow() {
     backgroundColor: '#00000000',
     ...appWindowIcon(),
     skipTaskbar: false,
-    ...(process.platform === 'darwin' && glass ? { vibrancy: 'hud', visualEffectState: 'active' } : {}),
+    ...(process.platform === 'darwin' ? { vibrancy: 'hud', visualEffectState: 'active' } : {}),
     ...(process.platform === 'win32' && glass ? { backgroundMaterial: 'acrylic' } : {}),
     webPreferences: {
       preload: path.join(__dirname, 'preload.js'),
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -1810,6 +1810,11 @@ function sendFloatingBubbleState() {
   try { mainWindow.webContents.send('floatingBubble:state', floatingBubblePayload()); } catch (_) {}
 }
 
+function sendMainWindowVisibility(win = mainWindow) {
+  if (!win || win !== mainWindow || win.isDestroyed() || win.webContents.isDestroyed()) return;
+  win.webContents.send('window:visibility', win.isVisible() && !win.isMinimized());
+}
+
 function stopFloatingBubbleAutoCollapseTimer() {
   if (floatingBubbleAutoCollapseTimer) clearTimeout(floatingBubbleAutoCollapseTimer);
   floatingBubbleAutoCollapseTimer = null;
@@ -5901,7 +5906,14 @@ function createWindow(boundsOverride, options = {}) {
     }
   });
   win.webContents.on('before-input-event', handleZoomShortcut);
-  win.webContents.once('did-finish-load', sendFloatingBubbleState);
+  win.on('show', () => sendMainWindowVisibility(win));
+  win.on('hide', () => sendMainWindowVisibility(win));
+  win.on('minimize', () => sendMainWindowVisibility(win));
+  win.on('restore', () => sendMainWindowVisibility(win));
+  win.webContents.once('did-finish-load', () => {
+    sendFloatingBubbleState();
+    sendMainWindowVisibility(win);
+  });
   loadWindowFile(win, {
     waitForContent: options.waitForContent === true,
     inactive: options.inactive === true,
@@ -5913,6 +5925,7 @@ function createWindow(boundsOverride, options = {}) {
         suppressInitialNumberAnimation: options.suppressInitialNumberAnimation === true,
         viewState: rendererViewState
       }),
+      ...(settings?.trayMode ? { windowHidden: '1' } : {}),
       ...(settings?.systemGlass === false ? { systemGlassDisabled: '1' } : {}),
       ...(windowsAccentFallback ? { windowsBackdropFallback: '1' } : {})
     }
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -324,6 +324,10 @@ const {
   normalizeWindowsBackdropMode
 } = require('./windowsBackdropMode');
 const { applyWindowsAccentBlur } = require('./windowsBackdrop');
+const {
+  attachNativeMaterialVisibility,
+  syncNativeMaterialVisibility
+} = require('./nativeMaterialVisibility');
 
 if (!app.isPackaged) loadDotEnv();
 
@@ -380,7 +384,9 @@ const DEFAULT_HOME_MODULE_LIST = ['limits', 'tool', 'device', 'model', 'trends']
 const TRAY_OPEN_VIEW_IDS = new Set(['home', 'project', 'session', 'limits', 'trends', 'status']);
 
 let mainWindow = null;
+let mainWindowNativeBlurEnabled = false;
 let dashboardWindow = null;
+let dashboardWindowNativeBlurEnabled = false;
 let settingsPath = null;
 let settings = null;
 let claudeWebCookieMutationRevision = 0;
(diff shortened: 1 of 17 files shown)


Javis603:
Thanks @REMIXPLAYER — the `powermetrics` work here is more careful than most perf PRs get, and the direction is right. I've pushed a follow-up commit to your branch instead of sending you round another cycle; please take a look and push back on anything you disagree with.

The main one: the glass was greying out whenever the window lost focus. `BrowserWindow.setVisualEffectState()` doesn't exist in Electron — verified on our pinned 43.4.0:

    PROBE {"electron":"43.4.0","hasSetVibrancy":"function","hasSetVisualEffectState":"undefined"}

`visualEffectState` is a constructor option only, so once it came off the constructor macOS fell back to `followWindow`. The `?.()` swallowed the missing method silently, and the fake window in the test *has* the method — so `states: ['active', 'inactive', 'inactive']` passed while the real behaviour never happened.

That trap is ours, not yours. `keepNativeBlurActive()` has called `setVisualEffectState` behind a `typeof === 'function'` guard since v0.1.0, so it has never once executed, while reading exactly as if the active state were being managed. It's what made me miss this on my first review pass too. Deleted now.

Three others, with the reasoning in the commit message:

- `did-finish-load` reported a not-yet-revealed window as hidden, which stopped the render that `waitForContent` reveals on — so floating-bubble collapse and expand fell back to the 2.5s timer on Windows. Forcing that path locally: 2501 ms → 186 ms, and via `contentReady` rather than the fallback.
- The header pin button stopped redrawing; `syncWindowBehaviorControls()` sat behind the closed-Settings guard. There's now a test asserting that nothing behind that guard writes main-surface DOM.
- A window revealed straight into Settings never signalled content ready, and `applyNativeMaterial()` re-applied an unchanged material to the Dashboard on every bubble collapse.

REMIXPLAYER:
Thanks @Javis603 for taking the time to review this so carefully — and for pushing the fixes straight onto the branch instead of sending me another review cycle. The `setVisualEffectState()` trap is exactly the kind of false green source-string tests can't catch, and the guard test closes that class. No pushback from my side.

Building on your baseline I'm verifying edge behaviours across the state matrix — hidden restore, first reveal, bubble/Settings transitions, glass re-enable, window replace — with real-Electron red tests. A few more transition gaps surfaced and are being fixed locally; I'll rebase onto your test-guard commit, push, and post the results here.

REMIXPLAYER:
**Where the added test weight came from**

Every test added after the first review maps 1:1 to a regression that actually fired during this PR's development — mostly in code that already had tests. Suspension changes fail at state transitions, not in steady state, and they fail in ways standard setups green-light:

| Trap | What now pins it |
|---|---|
| `setVisualEffectState()` doesn't exist; the guard + `?.()` never executed while the fake-window test passed | API-surface guard test + constructor assertion |
| `did-finish-load` reported a pre-reveal window as hidden → 2.5s timer fallback (2501 ms → 186 ms) | reveal-path assertions; M1 |
| Closed-Settings guard swallowed the pin button's writer; Settings-first reveal never signalled content ready | guard-boundary tests |
| Restore double-paint (`focus` while hidden → `visibilitychange`), found via persistent DOM counters | either-order coalescing test: both orders → one refresh |
| Hidden-phase races: in-flight `finally` re-issuing IPC, repeated invalidations refetching, session detail rebuilding DOM invisibly | queue-drain, stale-result, defer-repaint tests |
| First Home/Trends → Status switch read the previous frame's panel class; ticker never started | entering-Status test |
| Cursor's startup discovery rebuilt hidden Settings lists | surface-ownership rule: 11 exits self-guard, one test |

8 focused tests for ~10 distinct regressions — the count tracks the failures. Tests that only restated behaviour were deleted (the dead `flush()` stub), and your test we mistakenly removed was restored.

Scale: ~100 lines of production code, ~330 lines of tests, two packaged verification passes over the state coverage. Not claiming the suite is complete — only that every line was bought by a real failure, from exactly the failure class this PR can least afford.


Verification has since finished on the final commit stack: hidden-phase DOM writes measured at zero across repeated pushes, one catch-up repaint on restore, the tray icon refreshing while hidden, surface guards holding live, no renderer errors — per-row detail is in the description under State coverage.

Javis603:
Thanks @REMIXPLAYER for taking this beyond the original performance pass and closing out the transition matrix. The controlled `powermetrics` work was already thorough, and the added packaged-app/CDP verification, together with the 1:1 mapping from each observed regression to a guard or test, gives the final branch a much stronger validation story.

I reviewed the final `93971074` stack and exercised the macOS build with an isolated profile; the paths I tested held up without a visible regression. The exact-head suite passes here as well, with 3,788 passed and the one pre-existing skip. I’m happy with this final shape.

PR #389, opened 2026-08-12
feat(codex): estimate weekly quota value from local usage

lear111:
## Summary

Add a Codex weekly quota value estimate below the Codex weekly meter. It correlates locally observed official OpenAI Codex quota cost with continuously observed percentage transitions so users can sanity-check the approximate API-price value represented by their subscription quota.

Only `Estimated weekly quota` is shown; no local-device spend total is presented.

## User-visible behavior

- Show `Estimated weekly quota` directly below the Codex weekly limit.
- Show the estimate row for every identified Codex account; a newly added account starts in the collecting state.
- Restore each account's own persisted estimate for the matching weekly reset cycle without borrowing samples from the active account.
- Publish after three percentage points of valid continuously observed transitions, then update at each new boundary.
- Separate observations by Codex `accountKey` and weekly reset cycle.
- Preserve valid estimate evidence across updates and account switches.
- Normalize persisted segment state so an account switch cannot leave duplicate active segments.
- Start a new cycle after a natural or manual quota reset.
- Apply Codex Fast service-tier credit multipliers without changing ordinary API-cost totals elsewhere.

## Estimation model

Observations are grouped into uninterrupted local segments. The first usable percentage boundary in each segment is an anchor; later boundaries extend its endpoint.

```text
segment net cost = endpoint cumulative quota cost - anchor cumulative quota cost
segment span = endpoint used percentage - anchor used percentage

estimated weekly value =
  100 * sum(segment net costs) / sum(segment spans)
```

Temporary negative counter observations remain inside the segment, so a regression and its later rebound cancel through endpoint netting instead of creating another anchor.

A segment ends only when the active account changes, the weekly quota resets, or a directly observed percentage jump spans more than 1%. The first boundary in every new segment is normally an anchor. When the reset-cycle `100%` point was directly observed, it is instead retained as the exact endpoint-net start: `100% -> 99%` contributes to the accumulated span but cannot publish an estimate by itself, and `100% -> 97%` uses the total net cost across all three percentage points. A segment first observed below 100% still excludes its initial partial bucket. Jumps larger than 1% are retained for diagnostics but excluded from the formula.

Third-party or relay Codex sessions that cannot be attributed exclusively to OpenAI are excluded.

## Persistence and migration

State is stored locally in `codex-weekly-quota-samples.json`, grouped by hashed account key and quota cycle. Records contain aggregate percentages, token/cost counters, timestamps, segment identity, status, and diagnostic reason; they do not contain prompts, responses, credentials, or raw account identifiers.

Version 1 state migrates while retaining valid estimate evidence. Legacy current-device usage accumulators are discarded because transient counter regressions could inflate them.

Storage is bounded to 200 important observations and 32 segments per cycle, plus 12 cycles per account. Older closed segments are compacted into aggregate estimate offsets; old anomaly diagnostics are pruned before boundary records.

## Limitations

The calculation uses local official Codex session usage only. It cannot observe quota consumed on another device, and local token records do not contain an account id. Local counter intervals are therefore attributed to the `accountKey` reported as active by the Codex limits provider during that interval. An inactive newly added account can display its own saved estimate, but cannot collect new local cost evidence until it becomes the active local account.

## Validation

- Rebased onto current `main` and resolved the collector integration with the upstream Qoder CN adapter.
- `npm run lint` passed.
- Codex estimator, persistence, migration, pricing, multi-account selection, segment normalization, and presentation tests passed (21 focused tests).
- GitHub CI passed on Windows, macOS, Ubuntu Node 22/24, the macOS Widget build, and CodeQL.
- The full local suite completed 3,312 tests with 3,302 passing, 4 skipped, and 6 unrelated Windows permission, timing, and watcher-fixture failures.
- Windows NSIS and portable builds passed before the rebase; release artifact names were verified.

## Screenshot

The multi-account Codex panel keeps each account's estimate independent: one account restores a ready estimate while a newly added account starts in the collecting state.

Codex weekly quota estimate for multiple accounts


lear111:
Real-world Windows validation completed with the locally built x64 package.

- Codex weekly quota: 72% remaining
- Weekly quota estimate rendered successfully: **$127.68**
- The UI showed only the requested estimate amount, without normal/shrinkage classification
- The estimate appeared after the minimum safe observation span was reached

The local screenshot is being retained outside the repository because it contains user-specific usage data.

lear111:
Rebased this PR onto the current v0.47.0 mainline and resolved the collector integration conflicts. The refreshed branch now uses the upstream OAuth-first Codex quota path, preserves per-account historical samples, distinguishes reset-credit grants from actual weekly rollovers, and recovers estimates across cumulative-cost basis changes. The unrelated local game-interruption workaround is intentionally not included. Focused quota/presentation/store tests: 28/28 passing; the combined private build also passed 131/131 targeted Codex and limits-runtime tests.

lear111:
Follow-up: refreshed the Hub build registry and fixed one existing UTC-based Qoder history fixture exposed by the timezone matrix. The complete GitHub CI matrix now passes on Windows, macOS, Linux, all configured timezone offsets, CodeQL, vendored Tokscale verification, and the macOS Widget build.

lear111:
Included the witnessed 100%→99% bucket cost even when one or more same-percent 100% polls occur before the rounded boundary changes. Existing v2 ledgers recover automatically by normalizing the estimate start back to the witnessed reset-cycle endpoint. Added live-path and persisted-state regressions.

PR #413, opened 2026-08-14
fix(tray): attach the context menu with setContextMenu() on Linux

### LINKED ISSUE #417: Linux/GNOME: tray icon unusable — dead 3-dot icon (Electron 43 SNI regression) + context menu never attached + black icon invisible on dark panels
## Environment

- Token Monitor v0.44.0 AppImage (Electron 43.3.0), Linux x64
- Ubuntu 24.04, GNOME Shell 46, X11
- `ubuntu-appindicators@ubuntu.com` (Ubuntu's AppIndicator/KStatusNotifier extension)
- Other tray apps (Qt/Python ayatana) work fine on the same panel

There are three separate problems stacked on top of each other; the first two have a fix in #413, the third is an upstream Electron/Chromium regression that needs a maintainer decision.

## 1. The tray context menu is never reachable (app bug — fixed in #413)

`createTray()` only attaches the menu through the Electron `right-click` event + `tray.popUpContextMenu()`. Linux shells never deliver a `right-click` event — they render whatever menu the app exported over `com.canonical.dbusmenu` on right-click. Since `tray.setContextMenu()` was never called, the exported menu is empty:

```
$ dbus-send --session --print-reply \
  --dest=org.freedesktop.StatusNotifierItem-<pid>-1 /org/chromium/DbusMenu/1 \
  com.canonical.dbusmenu.GetLayout int32:0 int32:-1 array:string: dict:string:string:
   struct {
      int32 0
      array [ dict entry( string "children-display" variant string "submenu" ) ]
      array [ ]        <-- empty
   }
```

With `setContextMenu()` attached, `GetLayout` returns the full 125-item menu and right-click works.

Additionally, refreshing the menu must be signature-guarded: rebuilding/reattaching it on every stats tick re-exports the D-Bus menu hundreds of times per minute (I measured 400+ `LayoutUpdated` signals/min), which tray hosts treat as churn.

## 2. The tray icon renders as an invisible black square on dark panels (app bug — fixed in #413)

With `trayContent: "tokens"`, the renderer rasterizes a provider SVG (e.g. `assets/icons/opencode.svg`, authored with `fill="currentColor"` → black in canvas) or the macOS menubar template PNG (`tray-token-monitor.png`, solid black + alpha). macOS recolors template images; Linux tray hosts do not. Result on a dark GNOME panel: a fully black-on-black icon that looks like the tray is missing/dead. Verified by reading the exported icon pixmap — before: `(0,0,0,255)`, after fix: `(230,230,230,255)`.

## 3. Tray icon shows as dead 3-dot placeholder; clicks do nothing (upstream Electron 43 regression)

**This one is not fixable in app code as far as I could determine.**

On Electron 43, `Tray` registers its StatusNotifierItem using the well-known-name form (`org.freedesktop.StatusNotifierItem-<pid>-N`). GNOME's extension reads the item's properties immediately after `RegisterStatusNotifierItem` — and Chromium answers those reads with a D-Bus error, every time:

```
method call ... member=RegisterStatusNotifierItem
   string "org.freedesktop.StatusNotifierItem-442238-1"
...
method call ... member=GetAll            <-- from gnome-shell, ~2.5ms later
   string "org.kde.StatusNotifierItem"
error ... error_name=org.freedesktop.DBus.Error.Failed
   string "error occurred in GetAll"
method call ... member=Get  ("Id")      <-- extension retry, +0.8s
error ... "error occurred in Get"
method call ... member=Get  ("Menu")
error ... "error occurred in Get"
```

```
gnome-shell[...]: While initalizing proxy for org.freedesktop.StatusNotifierItem-442238-1:
  Gio.DBusError: ... Failed: error occurred in Get
ubuntu-appindicators@ubuntu.com[...]: ... Impossible to get basic properties
```

The extension then never builds the panel actor → the 3-dot placeholder icon that ignores left and right clicks. Notably:

- Direct `gdbus` property `Get`/`GetAll` calls to the same name/path **succeed** seconds later, so the item is alive — only the extension's immediate post-registration reads fail.
- `Activate` invoked directly over D-Bus works (window shows), confirming the app side is fine.

### Bisect / evidence

| Test | Result |
|---|---|
| 20-line minimal Electron 43 `Tray` app (no churn, menu attached) | same failure — dead actor |
| Minimal app deferring `new Tray()` until 25s after ready | same failure — not a startup-busy race |
| Recreating the Tray up to 5 times after launch | every registration fails the same way |
| Same minimal app on **Electron 37.2.0** | icon renders, right-click menu opens, all working |
| Non-Electron indicators (Qt CopyQ, Python ayatana) on the same panel | working |

On Electron 37, registration uses the `:unique-name@/StatusNotifierItem` form — same as the working Qt apps — while Electron 43 uses the well-known-name form; the property-read failure appears tied to that change.

### Suggested options

- Pin Linux builds to an Electron version whose tray still works (≤37 worked here; I did not bisect the exact breaking version), and/or
- Report upstream to electron/electron with the minimal repro + D-Bus traces above.

## Workaround for affected users

None found that keeps the tray functional on GNOME 46 + ubuntu-appindicators with Electron 43; the main window, global shortcut, and launcher entry still work.

lamualfa:
Closes #417

## Problem

On Linux, the Token Monitor tray was effectively unusable:
- right-clicking the tray icon did nothing (no context menu — Refresh, Open View, tray content mode, Quit all unreachable),
- the icon itself rendered as an invisible black square on dark panels.

(There is also a third, separate problem on Electron 43.3+ + GNOME — the dead 3-dot placeholder — which is an upstream Electron regression (electron/electron#52674, bisected to 84fe3290 / #52416; last good 43.2.0) and **not** addressed by this PR; see #417 §3. Pinning the Linux build to Electron 42.4.1/43.2.0 + this PR gives a fully working tray on GNOME.)

## Root cause

**Menu:** `createTray()` only attached the menu through the Electron `right-click` event + `tray.popUpContextMenu()`. Linux tray hosts implementing StatusNotifierItem (GNOME's AppIndicator/KStatusNotifier extension, KDE Plasma, Xfce status notifier, etc.) do not deliver a `right-click` event — the shell itself intercepts right-clicks and renders whatever menu the app exported over `com.canonical.dbusmenu`. Since `tray.setContextMenu()` was never called, the exported D-Bus menu was empty, so the shell showed nothing.

**Icon:** the renderer rasterizes provider SVGs authored with `fill="currentColor"` (→ black in canvas) and the macOS menubar template PNG (solid black + alpha). macOS recolors template images; Linux tray hosts do not, so on dark panels the glyph is black-on-black.

## Changes

- `src/electron/tray.js` — on Linux, attach the menu with `tray.setContextMenu()` so Electron exports it over D-Bus; keep `popUpContextMenu()` for Windows. Rebuild/reattach only when the menu-visible state changed (`trayMenuStateSignature`) — rebuilding on every stats tick re-exported the D-Bus menu 400+ times/minute (measured `LayoutUpdated` signal rate), which tray hosts treat as churn.
- `src/electron/main.js` — call `tray.refreshContextMenu()` from `updateTrayDisplay()`.
- `src/electron/renderer/app.js` — detect dark monochrome tray icon sources after rasterization and repaint them in a light neutral (`#e6e6e6`) on Linux only, when no explicit template colour was requested. Verified by reading the exported pixmap: before `(0,0,0,255)`, after `(230,230,230,255)`.

macOS is untouched, and the left-click `Activate` path is unchanged.

## Verification (GNOME 46, X11, ubuntu-appindicators, AppImage v0.44.0 with these patches)

- `com.canonical.dbusmenu.GetLayout` on the exported menu returns the full menu — Refresh Now, Open View → Home/…, Tray Display, Window Presentation, Version, Settings, Quit (0 items before, 125 after).
- Left-click (`Activate`) toggles/focuses the window.
- Icon stays visible on the dark panel instead of black-on-black.
- On an Electron 37 build, right-click opens the shell-rendered menu. (On Electron 43 the panel actor itself is broken upstream — see #417 §3 — so on GNOME 46 + ubuntu-appindicators the full click/menu path can't be exercised regardless of app code; the D-Bus layer above was verified directly.)
- `npm test`: 3076 pass, 0 fail.
- `npx eslint` on the touched files: clean.

### DIFF
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -2115,6 +2115,7 @@ function saveSettings(options = {}) {
       previousSettings
     });
     persistedSettingsSnapshot = cloneSettingsSnapshot(settings);
+    refreshTrayContextMenu();
     return true;
   } catch (error) {
     settings = previousSettings;
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -184,6 +184,7 @@ const {
   pickUsageTrayIconId,
   popoverBounds,
   reconcileCodexAccountSelection,
+  runTrayMenuAction,
   sortCodexAccountsForDisplay,
   shouldUseTemplateTrayIcon,
   trayShowsTitle
@@ -3721,11 +3722,16 @@ function updateDiscordRpcDisplay(stats) {
   updateDiscordRpc(stats, settings?.currency, compactTokenDisplayOptions());
 }
 
+function refreshTrayContextMenu() {
+  if (!tray || tray.isDestroyed()) return;
+  if (typeof tray.refreshContextMenu === 'function') tray.refreshContextMenu();
+}
+
 function updateTrayDisplay() {
   if (!tray || tray.isDestroyed()) return;
   // Keep the exported D-Bus menu in sync (radio checks, refresh state, Codex
   // accounts) — see the Linux note in createTray().
-  if (typeof tray.refreshContextMenu === 'function') tray.refreshContextMenu();
+  refreshTrayContextMenu();
   const visibleStats = electronPresentationStats(latestStats);
   const mode = settings?.trayContent || 'tokens';
   const currency = normalizeCurrency(settings?.currency);
@@ -4300,21 +4306,24 @@ function sendMainWindowEvent(channel, payload, isStillCurrent) {
 async function refreshFromTray() {
   if (trayRefreshInFlight) return;
   const widgetProducerOwner = captureMacWidgetProducerOwner();
-  trayRefreshInFlight = true;
-  try {
-    const stats = await fetchStats({ force: true });
-    // Collector ticks normally publish their own final snapshot. Only bridge the
-    // result when fetchStats returned a different object (for example, a remote hub
-    // fetch while an external headless agent owns collection).
-    if (stats && stats !== latestStats) {
-      sendPush({ event: 'stats', data: { stats, mode, reason: 'manual' } }, { widgetProducerOwner });
+  return runTrayMenuAction({
+    setInFlight: (value) => { trayRefreshInFlight = value; },
+    refreshContextMenu: refreshTrayContextMenu,
+    action: async () => {
+      try {
+        const stats = await fetchStats({ force: true });
+        // Collector ticks normally publish their own final snapshot. Only bridge the
+        // result when fetchStats returned a different object (for example, a remote hub
+        // fetch while an external headless agent owns collection).
+        if (stats && stats !== latestStats) {
+          sendPush({ event: 'stats', data: { stats, mode, reason: 'manual' } }, { widgetProducerOwner });
+        }
+      } catch (error) {
+        console.warn(`[tray] refresh failed: ${error.message}`);
+        showTrayRefreshError(error?.message || error);
+      }
     }
-  } catch (error) {
-    console.warn(`[tray] refresh failed: ${error.message}`);
-    showTrayRefreshError(error?.message || error);
-  } finally {
-    trayRefreshInFlight = false;
-  }
+  });
 }
 
 function setTrayContentFromMenu(value) {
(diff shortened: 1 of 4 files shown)


Javis603:
Thanks @lamualfa for tracing the Linux failure through GNOME's StatusNotifierItem and D-Bus path. The GNOME 46 AppImage and `GetLayout` verification made the root cause concrete, and `setContextMenu()` is the correct integration here.

I pushed follow-up commits 16e1e015, adc01638, and 9a5792ba to complete the persistent-menu lifecycle: refresh and Codex switching now publish both in-flight transitions, Window Presentation updates immediately, and successful settings or account saves invalidate the exported menu while unchanged states stay deduplicated.

PR #418, opened 2026-08-14
fix(pricing): fall back to tokscale catalog cache when offline

yixia1797:
## What

`tokscale pricing` refreshes its catalog over the network and only falls back to its local cache after 20-30s of network timeouts. The collector's pricing lookup budget is 3s, so on offline/restricted machines locally parsed clients (Proma) showed zero cost even when the price was already on disk.

This change reads tokscale's own catalog cache files (`~/.config/tokscale/cache/pricing-{litellm,openrouter,models-dev}.json`) directly when the pricing command fails — zero network, zero wait.

## Why

- The command is authoritative when it succeeds (fresher, better matching) and still runs first; the fallback only fires on timeout/offline failure.
- The fallback is an exact model-id lookup, so it can never substitute one model's price for another — worst case it stays cost-unavailable (the previous behaviour), never a wrong price.
- It parses once per file revision (size+mtime), so repeated ticks don't re-parse ~8.7K catalog entries.

## Scope

- `src/shared/collector.js`: local catalog cache parsing + fallback in `resolveModelPricing`
- `tests/shared/tokscalePricingCatalog.test.js`: 5 tests (parse, provider-prefix strip, malformed-file tolerance, offline fallback integration, command-still-wins)

No new dependencies; `collector.js` is not part of the Worker closure, so no `sync:worker`.

## Verification

`npm run verify` — lint clean, 3205/3206 tests pass.

---
## Summary by cubic
Falls back to `tokscale`’s local pricing catalog when the network lookup fails, so offline machines compute model costs within the 3s budget. Previously the command waited ~20–30s and returned zero cost; now we read `~/.config/tokscale/cache/pricing-*.json` immediately on failure.

- Aligns fallback rules with `tokscale`: exact keys win; bare ids use terminal-key matches only when all candidates publish identical rates; provider-scoped ids and routing labels (`auto`, `agent_review`) never borrow terminal matches.
- Discovers the catalog in canonical cache and legacy roots; honors `TOKSCALE_CONFIG_DIR` with canonical-first, hermetic discovery; on Windows, `HOME` overrides affect only the legacy home-rooted cache.
- Memoizes by file path/size/mtime and defers future-dated sources; any catalog change or eligibility shift invalidates the Proma pricing cache.
- Normalizes and validates rates; tolerates missing/malformed files and nulls; if no safe match, price stays unavailable. Command results still win when available.

<sup>Written for commit e425dc0a025e1ddf87259761e75632a8588b7bb6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/418?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/shared/collector.js
+++ b/src/shared/collector.js
@@ -17,7 +17,7 @@ const {
   deriveClientOverall
 } = require('./clientHealth');
 const { tokscalePackageNameForPlatform, tokscalePlatformKey } = require('./tokscalePlatform');
-const { customPricingPath } = require('./tokscaleConfig');
+const { customPricingPath, tokscaleConfigDir } = require('./tokscaleConfig');
 const {
   applyPeriodDelta,
   emptyPeriod,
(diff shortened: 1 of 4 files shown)


Javis603:
Thanks @yixia1797 for adding the offline pricing fallback for locally parsed Proma usage.

I pushed three follow-up commits (1049d11, 2236c3c, and e425dc0) to keep the disk fallback fail closed and aligned with Tokscale’s resolver and cache-loading semantics.

The final implementation keeps `tokscale pricing` authoritative, rejects ambiguous or non-model routing labels, validates the catalog schema strictly, and mirrors canonical and legacy cache discovery across platforms. The fallback remains scoped to Proma pricing, including WSL Proma.

PR #428, opened 2026-08-16
fix(history): derive the aggregate day boundary from producer calendars

DevMello:
## Problem

The first commit fixed the local-day default used by History normalization and streak calculations. The aggregate path had a separate boundary problem: `aggregateHistory()` runs in the Hub or Worker, while each History row and `periodWindows.today.key` are stamped from the producer's local calendar. Using the reader's wall clock could therefore re-key a fleet-wide daily window, and a future-dated producer key could become the lexical maximum and clear every device's daily tier while monthly totals still counted the same tokens.

## Fix

- Keep the original UTC-to-local correction for the collector stamp and local History defaults.
- Derive the aggregate boundary from the producers. Each device's `periodWindows.today` states a lower bound on its own current day: exactly `today.key` while the window is open, and at least the day after once `endsAt` has passed. The aggregate takes the latest such day. The reader's clock is a fallback for records that carry no window at all, never a term in that selection.
- Only devices that put days into the merge may move the boundary. The test is a non-empty daily tier rather than a `history` field being present, because `coerceHistory()` reduces a string, an array, a number and `{}` alike to an empty History.
- Validate the untrusted producer key as a real calendar day, and accept it only when some correct clock could be naming that day right now. A zone at offset `o` reads the UTC day of `now + o`, and every legal offset lies within `[-12h, +14h]`, so the admissible set is exactly the UTC days the 26-hour interval `[now - 12h, now + 14h]` touches — two calendar days for most of the UTC day, three only while the interval straddles two midnights. UTC is only a sanity bound here; it never chooses the aggregate day.
- Retain the local-clock fallback for pre-`periodWindows` agents as best effort, and read the clock once for the whole pass so a local midnight cannot fall between the fallback day and the instant the expiry and plausibility tests use.
- Keep the shared Node/Worker implementation and generated Worker registry in sync.

## Why these bounds and not the obvious ones

**Not `UTC day ± 1`.** It reads like the same rule and is not. A fixed `± 1` admits three calendar days at every instant, but the real envelope covers only two for 22 hours out of 24, and the extra day it admits is the future one for 10 of those hours. A device whose clock is one calendar day fast clears every other gate — its key is well formed, adjacent rather than absurd, and its `endsAt` is a day ahead too — so it wins the lexical maximum and takes the boundary. Read at 00:30 UTC with one such device idle on its own "today", the aggregate streak of every device that is on time went from 3 to 0.

**Not the reader's calendar day as the reference.** The ends of the offset range are 26 hours apart, so a UTC-12 reader and a UTC+14 producer are both correct and name days two apart. Measuring from the reader scores one real deployment as skew, and rejecting the only producer there does not merely lose a day: the fallback window ends before every row it holds, so the daily tier empties while the monthly tier still counts the same tokens.

**Not `max(readerClock, producerDay)`.** A hub east of a live producer would key the fleet on a day the only device with data has not reached — the original bug wearing the producer-derived design as a disguise.

**A closed window is not silence.** Dropping its key rather than advancing it let the clock fallback re-select the day the window had just declared finished. A UTC+14 laptop asleep one minute past its own midnight kept feeding a live streak to a UTC Worker for the next fourteen hours, while its own widget showed 0. Advancing it also retires a dormant fleet without a separate rule, since a long-closed window has a successor the plausibility bound already drops.

## Verification

- `node --test tests/shared/historyDayBoundary.test.js`: 20 passed
- `npm run verify`: 3313 tests, 3312 passed, 1 skipped
- `node --test tests/shared/hubBuild.test.js`: 13 passed
- Generated Worker shared-code parity: clean

The regression suite pins each of the cases above, plus malformed and impossible keys, every non-contributing record shape, history-less and explicitly disabled devices, dormant windows, the legacy local-clock fallback, and Worker behaviour parity. Every case fixes both the timezone and the instant, since the CI matrix runs UTC where the local and the UTC calendar day agree and these tests would otherwise pass against the bugs they exist to catch.

### DIFF
--- a/src/shared/collector.js
+++ b/src/shared/collector.js
@@ -29,7 +29,7 @@ const {
 } = require('./usage');
 const { collectWslUsage: collectWslUsageImpl, emptyWslBundle, probeWslState: probeWslStateImpl } = require('./wslUsage');
 const { hermesProfileWatchDirs, resolveHermesHome } = require('./hermesProfiles');
-const { mergeHistories, parseGraphResult, normalizeHistory } = require('./history');
+const { localDayKey, mergeHistories, parseGraphResult, normalizeHistory } = require('./history');
 const { retainDailyHistory, retainLiveDailyHistory } = require('./dailyHistoryArchive');
 const cursorAuth = require('./cursorAuth');
 const { findSessionFiles, codexSessionFile } = require('./sessionFiles');
@@ -511,12 +511,11 @@ function resetPromaPricingCache() {
   promaPricingCache.clear();
 }
 
-function localTodayKey(date = new Date()) {
-  const y = date.getFullYear();
-  const m = String(date.getMonth() + 1).padStart(2, '0');
-  const d = String(date.getDate()).padStart(2, '0');
-  return `${y}-${m}-${d}`;
-}
+// The collector's stamp and history.js's window/streak boundary have to be the same
+// calendar day or the aggregate re-keys what the collector wrote, so there is one
+// implementation rather than two formatters free to drift. Kept under the collector's
+// own name: it is exported and reads as the stamping side at its call sites.
+const localTodayKey = localDayKey;
 
 function collectionDate(now) {
   const value = typeof now === 'function' ? now() : now;
--- a/src/shared/history.js
+++ b/src/shared/history.js
@@ -522,7 +522,7 @@ function deviceHistoryRevision(devices) {
 }
 
 module.exports = {
-  num, sumTokens, parseGraphResult, computeIntensities, localDayKey,
+  num, sumTokens, parseGraphResult, computeIntensities, localDayKey, dayKeyAddDays,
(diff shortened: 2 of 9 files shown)


Javis603:
Thanks @DevMello for isolating the local-calendar boundary and following up on the Hub/Worker case.

I pushed follow-up commit d2e6fedd to derive the aggregate day boundary from producer calendars. The original UTC-to-local correction for the collector and local History defaults is unchanged; this follow-up fixes the separate aggregate path. The producer key remains the selected day, while UTC is used only as a plausibility envelope so a bad clock such as 2099 cannot become the lexical maximum and clear the fleet's daily tier. The envelope is anchored on UTC because comparing against the reader's local day incorrectly rejects the legitimate UTC-12/UTC+14 pair.

The PR description now reflects the final design, and the regression suite covers that opposite-end timezone case.

PR #433, opened 2026-08-16
fix(sync): preserve Hub drafts when changing upload frequency

ousugo:
## Summary

- Keep sync upload frequency auto-saving when changed.
- Preserve unsaved Hub URL, secret, device ID, and host port drafts across settings pushes.
- Use per-field edit revisions so in-flight saves cannot overwrite newer user edits.
- Add regression coverage for interval auto-save, rehydration, save races, and Host port drafts.

## Root cause

Changing the sync upload frequency saves only `syncUploadIntervalMs`, but the resulting settings push rehydrates the renderer form. The old form sync unconditionally restored persisted Hub values and overwrote unsaved drafts.

## User impact

Changing the sync upload frequency now takes effect immediately without discarding unsaved Hub fields. Hub fields remain local drafts until the user clicks Save.

If a field is edited while a Save is in flight, its newer value is preserved even when it temporarily matches the previously persisted value.

## Validation

- `npm run verify`
  - 3,322 passed
  - 1 skipped
  - 0 failed
- Focused renderer settings tests
  - 58 passed
- `git diff --check`

## Related issue

None — this regression was not previously reported.

### DIFF
--- a/src/electron/renderer/app.js
+++ b/src/electron/renderer/app.js
@@ -10940,7 +10940,8 @@ els.saveSettingsButton.addEventListener('click', async () => {
   const patch = {
     hubUrl: els.hubUrlInput.value.trim(),
     secret: els.secretInput.value,
-    deviceId: els.deviceIdInput.value.trim()
+    deviceId: els.deviceIdInput.value.trim(),
+    syncUploadIntervalMs: Number(els.syncUploadIntervalInput.value)
   };
   if (state.settings.hubMode === 'host') {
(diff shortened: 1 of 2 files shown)


Javis603:
Thanks @ousugo for identifying the regression and getting the fix started.

I pushed a couple of maintainer follow-ups to preserve the existing save semantics: `syncUploadIntervalMs` remains an independent auto-save, while unsaved Hub fields are now protected from settings rehydration and in-flight save races. Host port is covered by the same draft handling as well.

I also added regression coverage and updated the PR description to match the final implementation.

PR #434, opened 2026-08-17
fix(limits): keep the fetch timeout armed while the body is read

DevMello:
## Problem

`return response.json()` inside the `try` runs the `finally` at once, so `clearTimeout` fired before the body settled. The abort timer only bounded time to headers, and a response whose head arrives and whose body stalls fell through to the outer probe deadline instead.

## Fix

Await the body inside the `try` at the four sites that arm their own timer. Same shape `refreshClaudeAccessToken` and `grokLimits` already use. Providers wrapped in `runWithProbeDeadline` are unaffected, since that deadline already races the whole read.

## Verify

`npm run verify` green, failure set unchanged from main. Four regression tests stub a head then stall response, one per touched file. Each fails by timeout before the fix and passes after.

---
## Summary by cubic
Keep the per-request fetch timeout active while reading response bodies so head-then-stall responses abort promptly. Previously, clearing the timer in `finally` after `return response.json()` bounded only headers; now we `await response.json()` inside `try`, so stalled bodies hit the existing abort paths.

- Touches `src/shared/{copilotDeviceFlow,copilotLimits,limitCollector,minimaxLimits}.js`: change `return response.json()` to `return await response.json()` where a fetch timeout is armed. Providers wrapped by `runWithProbeDeadline` are unchanged.
- Behavior: Copilot sign-in now surfaces a `timedOut` error on a stalled body; limit fetchers report provider status `unavailable` in the same case. No API or interface changes.
- Tests: Add one regression per site that stubs a head-then-stall response; all fail before and pass after.

<sup>Written for commit 7ffa1b48ed4131af4919614b8fd8bf04777b0602. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/434?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/shared/copilotDeviceFlow.js
+++ b/src/shared/copilotDeviceFlow.js
@@ -96,7 +96,7 @@ async function fetchForm(url, body, deps = {}) {
       ...(controller ? { signal: controller.signal } : {})
     });
     if (!response.ok) throw errorWithStatus('unavailable', `${url} returned ${response.status}`);
-    return response.json();
+    return await response.json();
   } catch (error) {
     if (externalSignal?.aborted) throw errorWithStatus('cancelled', 'Sign-in cancelled');
     if (error?.name === 'AbortError') throw errorWithStatus('timedOut', 'GitHub request timed out');
--- a/src/shared/copilotLimits.js
+++ b/src/shared/copilotLimits.js
@@ -253,7 +253,7 @@ async function fetchJson(url, headers, deps = {}) {
           : 'unavailable';
       throw errorWithStatus(status, `${url} returned ${response.status}`);
     }
-    return response.json();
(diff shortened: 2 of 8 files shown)


Javis603:
Thanks @DevMello for catching this. Keeping the timeout armed until response.json() settles is the right fix for the head-then-stall case, and the regression tests cover it well.

PR #441, opened 2026-08-17
fix(settings): stop the raw patch overwriting its own normalization

DevMello:
## Problem

`settings:update` applies about fifty clamps, enum guards and fallbacks over the spread patch, then hands the raw patch to `normalizeWindowBehaviorSettings`, which does `{ ...settings, ...patch }` internally. The raw values win a second time.

```js
normalizeWindowBehaviorSettings(
  { deviceId: 'my-box', glassOpacity: 68, refreshMs: 15000, hubHostPort: 17321 },
  { deviceId: '   ',    glassOpacity: '9999', refreshMs: 1,  hubHostPort: 70000 })

// -> { deviceId: '   ', glassOpacity: '9999', refreshMs: 1, hubHostPort: 70000 }
```

Clearing the Device ID field is the reachable case. Its placeholder offers `auto (hostname)` and the fallback implementing that is one of the discarded lines, so an empty `deviceId` is persisted and the widget stops recognising its own device in client and host mode.

## Fix

Narrow the patch to the two keys that select a window mode. The first spread has already applied every patch key, so nothing else needs the second pass. The helper sits next to the code that reads those keys so the list has one owner.

## Verify

`npm run verify` green, failure set unchanged from main. Reverting just the call site fails the new source assertion.

---
## Summary by cubic
Stops settings:update from reapplying raw, unnormalized values by narrowing the second merge to only window mode selectors. Previously, normalizeWindowBehaviorSettings spread the raw patch again and reinstated invalid values (e.g., empty deviceId persisted); now, only windowBehavior and alwaysOnTop are considered, so prior normalization stands.

| Area | Before | After |
| --- | --- | --- |
| Patch merge in window behavior normalization | Raw patch was spread over settings again, reintroducing unnormalized values | Only `windowBehavior` and `alwaysOnTop` are merged; other normalized values remain intact |
| Device ID fallback | Empty `deviceId` was persisted, breaking device recognition | Fallback to hostname applies; device remains recognized |
| Clamps/guards (e.g., `refreshMs`, port) | Floors/guards were overridden by raw inputs | Floors/guards are preserved after update |

**Implementation and tests**
- Introduces windowBehaviorSelection(patch) to keep only `windowBehavior` and `alwaysOnTop`, and uses it when calling normalizeWindowBehaviorSettings.
- Adds tests to verify selection behavior, preservation of normalized values, and the call site change in main.js.
- No migrations; risk limited to settings update path.

<sup>Written for commit 19ce3fbacb018d1a7ee52be220d0cb0618493183. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/441?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -279,7 +279,7 @@ const {
   runManualDeviceRefresh,
   settingsLimitInvalidationPlan
 } = require('./deviceRuntimeCoordinator');
-const { describeWindowBehavior, normalizeWindowBehaviorSettings } = require('./windowBehavior');
+const { describeWindowBehavior, normalizeWindowBehaviorSettings, windowBehaviorSelection } = require('./windowBehavior');
 const {
   normalizeWindowToggleShortcut,
   windowToggleShortcutAction,
@@ -6119,7 +6119,7 @@ app.whenReady().then(() => {
       customModelPricing: patch.customModelPricing !== undefined
         ? normalizeCustomPricingSetting(patch.customModelPricing)
         : normalizeCustomPricingSetting(settings.customModelPricing)
-    }, normalizedPatch);
+    }, windowBehaviorSelection(normalizedPatch));
     settings.archivedClientUsage = normalizeArchivedClientUsage(settings.archivedClientUsage);
     if (settings.clients !== previousClients) updateArchivedClientUsage(previousClients, settings.clients);
     delete settings.edgeDrawerEnabled;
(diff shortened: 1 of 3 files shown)


Javis603:
Thanks @DevMello for catching the double merge — verified against current `main`, `npm run verify` green, merging.

PR #443, opened 2026-08-17
feat(limits): hide unconfigured providers on the limits page

Excelius-Wang:
## Summary

The limits page painted a "Not set up" card for every enabled provider, including ones the user never configured. This hides those empty setup rows by default.

Detection is the row's own status plus whether it has quota data (`windows` / `balance` / `balanceUsd`), not a provider allowlist. OpenCode and Claude with ambient credentials still show when they arrive as `ok`; unauthorized and error rows stay visible. Setup remains in Settings.

This is display-only: enabled providers are still probed, so a later CLI login shows up on its own. Home already drops accounts with no displayable windows and is unchanged. Settings → AI Tool Limits has an opt-out checkbox to restore the old list.

The issue also asked not to fetch unused status-page news. That is already the Home → Status provider list; this PR does not change it.

## Test plan

- [x] `npm run verify`
- [ ] Limits page: enabled but unconfigured providers (e.g. DeepSeek with no API key) no longer render empty cards
- [ ] A configured provider, a login failure, and a probe error still render
- [ ] OpenCode / Claude with a local login still appear without extra Token Monitor setup
- [ ] Unchecking "Hide providers that are not set up" restores the previous full list
- [ ] Unchecking a provider in the Settings list still stops probing, as before
- [ ] Home limit module is unchanged

---
## Summary by cubic
Hide unconfigured providers on the Limits page to reduce noise while keeping failures and any provider with quota data visible.

- Hides rows with status `notConfigured`/`noSyncedData` that also have no `windows`/`balance`/`balanceUsd`.
- Ambient-credential providers still show as `ok`; `unauthorized`/`error`/`unavailable` rows remain visible.
- Display-only change: probing of enabled providers and Home limits module are unchanged.
- Adds an opt-out setting `hideUnconfiguredLimitProviders` (default on) in Settings → AI Tool Limits.

| Area | Before | After |
| --- | --- | --- |
| Limits page visibility | Showed a “Not set up” card for every enabled provider | Hides only `notConfigured`/`noSyncedData` rows with no quota data |
| Settings default | No toggle for this behavior | New “Hide providers that are not set up” toggle (on by default) |

**Notes for reviewers**
- Detection is data-driven (status + presence of quota), no provider allowlist; rendering only includes visible rows.
- New helpers `limitProviderHasQuota` and `shouldShowLimitProviderRow`; `renderLimits` filters rows and updates its render signature.
- Tests added for settings plumbing, presentation logic, and render behavior; `npm run verify` passes.
- Risk: Users expecting to see all enabled providers may miss hidden ones; uncheck the new setting to restore the full list.

<sup>Written for commit 38fca18b4e2ca4b141d5c276fb965cc86694e73d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/443?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Javis603:
Thanks for working on this. This does address #54 as originally described, but after thinking through the behavior more, I don't think we want enabled providers to be hidden dynamically based on their runtime configuration or data state.

I'd rather keep provider visibility entirely user-controlled after setup, and address the initial clutter only on a true first run by seeding the initial Limits provider selection from the AI tools already detected by the collector.

I've opened #537 to track that direction in more detail. Since this is a product-direction change rather than an implementation issue with this PR, I'm going to close this rather than ask you to rework it.

Thanks again for taking this on.

PR #482, opened 2026-08-21
test: make local-day fixtures timezone-independent

### LINKED ISSUE #481: npm run verify is only green at some UTC offsets: day-boundary fixtures are written in UTC while the code buckets by local midnight
## Summary

`npm run verify` — the single local entry point named in `AGENTS.md` — is only green at a handful of UTC offsets. On a clean checkout of `1519ecf` with `npm ci` and nothing else, the suite fails in several real timezones, and which tests fail changes with the offset.

The product code is not wrong. `buildQoderCnPeriods` and friends bucket by **local** midnight, which is what a desktop usage widget should do:

```js
// src/shared/qoderCnUsage.js:476
const todayStart = new Date(now.getFullYear(), now.getMonth(), now.getDate()).getTime();
```

The fixtures that exercise those boundaries are written as absolute UTC instants (`Date.parse('2026-07-29T18:00:00Z')`, rows at `now - 2h`). At UTC+00 both rows land on the same local day and the assertion holds. At UTC+07 the same two instants straddle local midnight, so `periods.today.totalInput` is `0` instead of `20`.

## Reproduce

```bash
git clone --depth 1 https://github.com/Javis603/token-monitor.git
cd token-monitor && npm ci
TZ=UTC              npm test   # pass 3534, fail 0
TZ=Asia/Ho_Chi_Minh npm test   # pass 3532, fail 2
```

Node 25.9.0 on macOS 15 (arm64). `npm run lint` is clean in every case; only `npm test` moves.

## Measured

Same checkout, same command, only `TZ` changed:

| `TZ` | offset | pass | fail |
|---|---|---|---|
| `UTC` | +00 | 3534 | 0 |
| `Europe/Berlin` | +02 | 3534 | 0 |
| `Australia/Sydney` | +10 | 3534 | 0 |
| `Asia/Ho_Chi_Minh` | +07 | 3532 | **2** |
| `America/Los_Angeles` | −07 | 3531 | **3** |
| `Pacific/Honolulu` | −10 | 3528 | **6** |
| `Pacific/Kiritimati` | +14 | 3528 | **6** |

It is not "UTC vs everything else" — +02 and +10 are green and +07 is not. Each fixture crosses a local boundary at its own set of offsets, so the failure set is different in each timezone rather than being one broken test:

- `Asia/Ho_Chi_Minh`: `tests/shared/qoderCnUsage.test.js:79`, `:181`
- `America/Los_Angeles`: `tests/shared/reasonixSessions.test.js:202`, `tests/shared/sessionDetail.test.js:170`, `tests/shared/sessionUsageArchive.test.js:215`
- `Pacific/Honolulu`: the three above plus `tests/shared/macWidgetSnapshot.test.js:80`, `tests/shared/reasonixSessionDetail.test.js:130`, `tests/shared/reasonixSessions.test.js:164`
- `Pacific/Kiritimati`: `tests/shared/clientUsageArchive.test.js:165`, `tests/shared/promaUsage.test.js:188`, `tests/shared/reasonixSessionDetail.test.js:54`, `:130`, `tests/shared/reasonixSessions.test.js:202`, `tests/shared/sessionDetail.test.js:170`

Eight files across the union.

## Why CI cannot see this

`.github/workflows/ci.yml` varies two axes — OS (`ubuntu-latest`, `macos-latest`, `windows-latest`) and Node (`22`, `24`) — and every GitHub-hosted runner has its clock set to UTC. So all five legs sample the same single offset, and an offset-dependent assertion is exactly the thing a matrix like this cannot reach. The gate is green by construction, and the first person to see the red is a new contributor running the documented command on their own machine.

## Sample failure

```
✖ buildQoderCnPeriods keeps day boundaries and tokscale-compatible totals
  AssertionError: Expected values to be strictly equal:
  0 !== 20
      at tests/shared/qoderCnUsage.test.js:86:10
```

`now = 2026-07-29T18:00:00Z`; the second row is `now - 2h` = `16:00Z`. At UTC+07 that is `2026-07-29 23:00` local while `now` is already `2026-07-30 01:00` local, so the row is correctly excluded from "today" — and the assertion, written for UTC, expects it to be included.

## Two ways to fix, and they are not equivalent

1. **Pin the suite's timezone.** Cheapest, and it makes the gate deterministic for every contributor. But local midnight then always equals UTC midnight, so the day-boundary logic is only ever exercised at one offset and the class of bug the tests exist to catch stays untestable. It also needs care to stay cross-platform, since the Windows leg cannot take a `TZ=… npm test` prefix.
2. **Anchor the fixtures to local midnight** instead of to a `Z` literal — build `now` from `new Date(y, m, d, 12, 0, 0)` and derive the rows from it, so "two hours ago" is two hours ago in whatever timezone the suite is running. The tests then assert the same property at every offset, and a future regression in local-day bucketing fails everywhere rather than nowhere.

The second is the one that keeps the coverage, and it is mechanical per fixture. Adding a second `TZ` to the CI matrix afterwards is what stops it coming back.

Happy to send the PR. Would you prefer one PR across the eight files, or one per file so each stays reviewable on its own?

vinhnguyenthanhdn:
## Summary

Makes the local-day and local-month test fixtures reported in #481 timezone-independent while preserving their intended boundary coverage.

- Adds shared local-time fixture helpers in `tests/helpers/localTime.js`.
- Converts the affected Qoder CN, Proma, Reasonix, session archive/detail, client archive, and macOS Widget fixtures to use local calendar parts.
- Keeps the boundary-sensitive cases immediately before and after local midnight.
- Adds full-suite CI coverage across representative positive, negative, extreme, and fractional timezone offsets.

No runtime code or API behavior changed.

## Coverage

The fixtures now express the local calendar day they intend to test instead of relying on UTC `Z` literals. Mutating the production logic to cut periods at UTC midnight is caught outside UTC, so the change removes timezone-dependent failures without weakening local-midnight coverage.

## Validation

- `npm run verify`: 3577 passed, 0 failed, 1 skipped
- All GitHub CI jobs passed, including the timezone and macOS Widget jobs
- Proma tests passed across UTC and all seven additional CI timezones

Closes #481

---
## Summary by cubic
Pins day/month boundary and trend fixtures to local time to match production bucketing and prevent false failures at non-UTC offsets; expands CI to sample negative and fractional timezones.

- Adds `tests/helpers/localTime.js` (`localDate`, `localMs`, `localIso`) and converts fixtures in `qoderCnUsage`, `promaUsage`, `macWidgetSnapshot`, `reasonixSessionDetail`, `reasonixSessions`, `sessionDetail`, `clientUsageArchive`, and `sessionUsageArchive` to local time.
- Anchors select fixtures one minute on either side of local midnight to assert boundaries.
- Adds a `timezones` job in `.github/workflows/ci.yml` to run tests under several `TZ` values, including fractional and negative offsets.

| Area | Before | After |
| --- | --- | --- |
| Test fixture clocks | UTC instants that silently shift calendar days at different offsets | Instants derived from local calendar parts via `localTime` helpers |
| Boundary coverage | Some boundary assertions only held at UTC; UTC-midnight regression could pass | Fixtures straddle local midnight; UTC-midnight regression fails across offsets (except UTC) |
| CI timezone sampling | All jobs ran at UTC only | `timezones` job runs the suite at several offsets (+07, +05:45, +12:45, -07/-08, -10, -11, +14) |

- Implementation: runtime unchanged; centralize timezone reasoning in `tests/helpers/localTime.js`.
- Validation: suite passes across listed `TZ`s; mutants that cut days at UTC midnight now fail outside UTC.
- Migration: when asserting day/month windows, use `localDate`/`localMs`/`localIso`; avoid `Z` literals for boundary-sensitive cases.
- Risk: slightly longer CI from the added job; no runtime or API changes.

<sup>Written for commit 0b862d9edaa60d11d67323858ed9587326fd898b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/482?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
       - run: npm ci
       - name: Run the suite at several UTC offsets
         run: |
-          for tz in Asia/Ho_Chi_Minh Asia/Kathmandu America/Los_Angeles Pacific/Honolulu Pacific/Kiritimati; do
+          for tz in Asia/Ho_Chi_Minh Asia/Kathmandu Pacific/Chatham America/Los_Angeles Pacific/Honolulu Pacific/Pago_Pago Pacific/Kiritimati; do
             echo "::group::TZ=$tz"
             TZ="$tz" npm test
             echo "::endgroup::"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,32 @@ jobs:
       - run: npm run lint
       - run: npm test
 
+  timezones:
+    # Every GitHub-hosted runner sets its clock to UTC on every OS, so the
+    # os × node matrix above varies two axes and still samples a single UTC
+    # offset. Usage periods are cut at *local* midnight, so an assertion written
+    # against a UTC instant is green there by construction and the first person
+    # to see it red is a contributor running `npm run verify` at home.
+    #
+    # This is a list rather than "UTC plus one other" because each offset failed
+    # a different subset of the suite (#481): +02 and +10 were green while +07
+    # was not, and Kathmandu is here because +05:45 is not a whole hour.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '24'
+          cache: npm
+      - run: npm ci
+      - name: Run the suite at several UTC offsets
+        run: |
+          for tz in Asia/Ho_Chi_Minh Asia/Kathmandu America/Los_Angeles Pacific/Honolulu Pacific/Kiritimati; do
+            echo "::group::TZ=$tz"
+            TZ="$tz" npm test
+            echo "::endgroup::"
+          done
+
   mac-widget:
     runs-on: macos-latest
     steps:
--- a/tests/helpers/localTime.js
+++ b/tests/helpers/localTime.js
@@ -1,0 +1,29 @@
+'use strict';
+
+// Fixtures that exercise day or month bucketing have to state their instants in
+// the same clock the product code buckets by, and that clock is local: usage
+// periods are cut at local midnight (`src/shared/qoderCnUsage.js`, the archive
+// modules, the widget trend), because a desktop widget belongs to the calendar
+// the person reading it lives in.
+//
+// A `Z` literal only lands on the calendar day it names at some offsets. Written
+// as `2026-07-17T08:30:00.000Z` it is the 17th at UTC+00 and the 16th at UTC-10,
+// so an assertion built on it is green at some offsets and red at others — and
+// every GitHub-hosted runner sits at UTC+00, so the OS × Node matrix samples
+// exactly the one offset where the mismatch is invisible.
+//
+// Building the instant from local calendar parts states the day the fixture
+// means. The assertion then holds at every offset, and a regression that cut the
+// day at UTC midnight fails everywhere except UTC instead of passing everywhere
+// except a few offsets.
+//
+// `month` is 1-based so a call transcribes the ISO literal it replaces
+// (`2026-07-17` → `2026, 7, 17`), not `Date`'s 0-based monthIndex.
+const localDate = (year, month, day, hour = 0, minute = 0, second = 0, ms = 0) =>
+  new Date(year, month - 1, day, hour, minute, second, ms);
+
+const localMs = (...parts) => localDate(...parts).getTime();
+
+const localIso = (...parts) => localDate(...parts).toISOString();
+
+module.exports = { localDate, localMs, localIso };
--- a/tests/shared/clientUsageArchive.test.js
+++ b/tests/shared/clientUsageArchive.test.js
@@ -14,6 +14,8 @@ const {
   pruneArchivedClientUsage
 } = archiveApi;
 
+const { localDate } = require('../helpers/localTime');
+
 function deviceRecord() {
   return {
     deviceId: 'macbook',
@@ -163,11 +165,15 @@ test('archived client usage is added back while the client remains untracked', (
 });
 
 test('archived day and month usage follow calendar boundaries', () => {
-  const archive = captureArchivedClientUsage({}, deviceRecord(), ['hermes'], new Date('2026-05-30T12:00:00.000Z'));
+  // The day and month windows are cut at local midnight, so the three clocks
+  // below are stated in local time: a `Z` noon is already the next calendar day
+  // past UTC+12, which walks the capture and both reads a day forward together
+  // and takes the month rollover with them.
+  const archive = captureArchivedClientUsage({}, deviceRecord(), ['hermes'], localDate(2026, 5, 30, 12));
 
   const nextDay = applyArchivedClientUsage(liveSummaryWithoutHermes(), archive, {
     activeClients: 'codex',
-    now: new Date('2026-05-31T12:00:00.000Z')
+    now: localDate(2026, 5, 31, 12)
   });
   assert.equal(nextDay.today.clients.hermes, undefined);
   assert.equal(nextDay.today.models['claude-3-5-sonnet'], undefined);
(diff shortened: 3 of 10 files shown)


Javis603:
The fixture change in this PR is correct and directly exercises the local-midnight boundary.

I would prefer one PR covering all eight fixtures rather than one PR per file. Please add the remaining seven files to #482. A lightweight non-UTC CI check would also help prevent this class of regression from returning.

Thanks!

vinhnguyenthanhdn:
One PR it is — the remaining seven files are in, along with the non-UTC check. `0151359`.

The eight fixtures now build their instants from local calendar parts through a small `tests/helpers/localTime.js`, which is where the reasoning sits so the next fixture author reads it once rather than rediscovering it. Two of them are anchored a minute either side of local midnight instead of being pulled away from the boundary, which is the difference between a fixture that is green everywhere and one that still measures what it was written for.

Measured on this branch, same checkout, only `TZ` changed:

| `TZ` | offset | pass | fail |
|---|---|---|---|
| `UTC` | +00 | 3577 | 0 |
| `Europe/Berlin` | +02 | 3577 | 0 |
| `Asia/Kathmandu` | +05:45 | 3577 | 0 |
| `Asia/Ho_Chi_Minh` | +07 | 3577 | 0 |
| `Australia/Sydney` | +10 | 3577 | 0 |
| `America/Los_Angeles` | −07 | 3577 | 0 |
| `Pacific/Honolulu` | −10 | 3577 | 0 |
| `Pacific/Kiritimati` | +14 | 3577 | 0 |

`npm run lint` is clean, and no file under `src/` is touched.

Green everywhere is the weaker half of the claim, so here is the other half. Mutating the product code to cut the day and month at UTC midnight instead of local leaves **9 tests failing at UTC and 20 at +14** — and `qoderCnUsage.test.js:81`, `promaUsage.test.js:189` and `macWidgetSnapshot.test.js:86` are among the ones that die only *outside* UTC, which is exactly the coverage that did not exist before. Same for `clientUsageArchive.test.js:167` and `sessionUsageArchive.test.js:217` under the matching mutation in the archive modules. The rest of the converted fixtures assert day *separation* rather than the boundary position, so they hold in either clock by construction.

Two things fell out of the conversion that were not in #481:

- **`macWidgetSnapshot`'s fingerprint test was mixing clocks.** It built `first` from a `Z` literal and every comparison from the file's shared `NOW`. The claim is "the clock does not reach the fingerprint", but the pair was also stepping the trend's local day, so the test passed at UTC for a reason unrelated to what it says. It now uses two clocks an hour apart on one local day.
- **`reasonixSessions` already had the right idiom.** `localDateIso` is used by most of that file; the two failing tests simply had not been written against it. So this is less a new convention than the one already there, applied consistently — which is part of why a helper felt right over eight local copies.

On the CI side, the new `timezones` job runs the **whole suite** at five offsets rather than the files this PR touches. Naming those files would be the same shape of gap as the fixtures it is meant to guard: it would pass the day someone adds a ninth day-boundary fixture elsewhere. Five full runs cost well under two minutes on `ubuntu-latest`. `Asia/Kathmandu` is in the list because +05:45 is not a whole hour, and `Europe/Berlin`/`Australia/Sydney` are left out on purpose — they were already green in #481, so they buy nothing the existing UTC matrix does not.

One thing I deliberately did not do: pin `TZ` for the main matrix. That would make the default gate deterministic, but it also fixes local midnight at UTC midnight forever, and then this class of bug is untestable rather than merely unobserved. Happy to add it as well if you would rather have the determinism — it is a one-line `env:` on the `test` job.

Javis603:
Thanks @vinhnguyenthanhdn for the thorough timezone investigation, and for expanding the fix across all eight reported fixtures with full-suite timezone coverage. Moving the shared reasoning into `tests/helpers/localTime.js` keeps the intended local-day semantics explicit, while the midnight boundary cases retain the coverage they were meant to provide.

I pushed a small follow-up commit 0b862d9 to convert the two remaining Proma fixtures and add `Pacific/Pago_Pago` and `Pacific/Chatham` to CI. The final branch keeps runtime behavior unchanged and covers the negative and fractional offsets we verified.

PR #485, opened 2026-08-22
fix(grok): inject the macOS system proxy into the CLI child env

XiaoZ-0218:
## Problem

On macOS with a system-level proxy (ClashX / Surge / corporate proxy), the Grok limit card shows `unavailable` and later `unauthorized`, even though Grok usage tracking itself works and the same account reads fine from a shell.

Two stacked failures, both reproduced on v0.47.0:

1. **Billing RPC child cannot reach the network.** The widget's own fetch follows the OS proxy via Electron `net.fetch` (#380), but the Grok CLI spawned for the `agent stdio` x.ai/billing RPC is a plain Node program that only honors proxy **env vars** — and a GUI-launched app never has any. The child's requests to grok.com/x.ai fail, so the RPC path always errors and the card falls back (or fails) with no network path at all.
2. **Silent OAuth token expiry.** The CLI is also what refreshes `~/.grok/auth.json` (6h access tokens). With no proxy in the child env, the refresh fails silently. The web fallback then reads an expired token → `unauthorized`. Practical symptom: the card works right after any shell-side `grok` use (shells do have the proxy env), then dies exactly one token lifetime later, which looks like random flapping.

## Fix

Give spawned CLIs the same OS-proxy reach the widget's own fetch already has:

- **`src/shared/systemProxyEnv.js`** (new): reads `scutil --proxy` once in the background and exposes a *synchronous* `withSystemProxyEnv(env)` that injects `HTTPS_PROXY` / `HTTP_PROXY` / `ALL_PROXY` (both casings) plus `NO_PROXY` passthrough. Explicit env proxy vars always win untouched — same precedence as `createOutboundFetch`. Non-darwin platforms are untouched. scutil failures degrade to no injection (today's behavior), never to a rejected probe.
- **`src/shared/grokLimits.js`**: spawns the CLI with that env. The prime happens at `LimitsRuntime.start()` (fire-and-forget) so the probe path never waits on scutil; the first probe before the prime completes runs exactly as it does today.
- Tests: scutil parsing (HTTPS/HTTP/SOCKS precedence, all-disabled), env precedence (uppercase/lowercase/NO_PROXY), prime-then-inject lifecycle, spawn-env capture through `fetchGrokRpcBilling`, opt-out flag, and no-injection-before-prime.

## Verification

- Full suite: 3579 tests pass (`npm run lint && npm test`), including the 43 existing grokLimits tests unchanged.
- Live-reproduced the original failure on macOS + ClashX before the fix (RPC child offline → `unavailable`, token refresh dead → `unauthorized` after 6h), and verified the injected env lets the CLI both answer billing and refresh the token.

Caveat honestly stated: the cache lives for the process lifetime, so toggling the system proxy requires an app restart to re-read. That matches the widget's own restart cadence and degrades to today's behavior, but reviewers may prefer a TTL — happy to add one if you'd rather.

Reproduced and verified on macOS 15 (arm64), Token Monitor v0.47.0, ClashX system proxy.

---
## Summary by cubic
Injects the macOS system proxy into the Grok CLI child env so GUI-launched widgets can reach x.ai behind a system proxy. This restores billing RPC connectivity and token refresh on proxied networks.

- Adds a small helper to read `scutil --proxy`, cache it, and synchronously inject `HTTPS_PROXY`/`HTTP_PROXY`/`ALL_PROXY` (both casings) and preserve `NO_PROXY`.
- Spawns the Grok CLI with the merged env; explicit env proxies still win; prime runs in the background at runtime start; non-darwin platforms are unchanged.

| Area | Before | After |
| --- | --- | --- |
| Grok CLI network on macOS with system proxy | Child had no proxy env; requests to grok.com/x.ai failed | Child inherits OS proxy via injected HTTPS_PROXY/HTTP_PROXY/ALL_PROXY; RPC reaches network |
| Token refresh | CLI failed to refresh ~/.grok/auth.json; card becomes unauthorized after ~6h | Refresh works through proxy; card stays authorized |
| Env proxy precedence | Only explicit env proxies, if present, were used; GUI launch rarely had any | Explicit env still wins; OS proxy used only as fallback; NO_PROXY preserved |
| Probe latency at startup | None, but no proxy injection | OS proxy lookup primed in background; first probe may run unproxied; subsequent probes use cache |

**Notes for reviewers**
- Caches the system proxy for the process lifetime; changing the system proxy requires an app restart to re-read.
- On `scutil` failure or disabled proxies, no injection occurs (degrades to current behavior).
- Tests cover `scutil` parsing, env precedence (upper/lower), injection lifecycle and opt-out, and spawn-env capture via the billing RPC path.

<sup>Written for commit 1bdfd0eb5751bf5b98caa3b61f606a0fd260b9bc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/485?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/shared/grokLimits.js
+++ b/src/shared/grokLimits.js
@@ -17,6 +17,7 @@ const { spawn } = require('node:child_process');
 const { normalizeLimitProvider } = require('./limits');
 const { hashKey } = require('./hashKey');
 const { createOutboundFetch } = require('./outboundFetch');
+const { withSystemProxyEnv } = require('./systemProxyEnv');
 const { abortError } = require('./probeDeadline');
 
 const GROK_WEB_BILLING_GRPC_URL = 'https://grok.com/grok_api_v2.GrokBuildBilling/GetGrokCreditsConfig';
@@ -440,6 +441,19 @@ async function fetchGrokRpcBilling(options = {}, deps = {}) {
   const timeoutMs = Number(deps.rpcTimeoutMs || deps.fetchTimeoutMs || 5000);
   const signal = deps.signal;
   if (signal?.aborted) throw abortError(signal);
+  // The CLI is a plain Node program: it only reaches the network through
+  // proxy env vars, which a GUI-launched widget never has. Give it the OS
+  // proxy (primed in the background at startup — see systemProxyEnv.js) so
+  // both the billing RPC and its ~/.grok/auth.json token refresh can succeed.
+  // Existing env proxy vars always win; the first probe before the prime
+  // completes simply runs as it would today.
+  const childEnv = deps.systemProxyEnv === false
+    ? env
+    : withSystemProxyEnv(env, {
+      cachedSystemProxy: typeof deps.cachedSystemProxy === 'string'
+        ? deps.cachedSystemProxy
+        : undefined
+    });
 
   return new Promise((resolve, reject) => {
     let child;
@@ -510,7 +524,7 @@ async function fetchGrokRpcBilling(options = {}, deps = {}) {
 
     try {
       child = spawnFn(command, args, {
-        env,
+        env: childEnv,
         stdio: ['pipe', 'pipe', 'pipe']
       });
     } catch (error) {
--- a/src/shared/limitsRuntime.js
+++ b/src/shared/limitsRuntime.js
@@ -31,6 +31,7 @@ const {
   computeRetryDelayMs,
   isRetryableLimitStatus
 } = require('./limitsRetryPolicy');
+const { primeMacSystemProxy } = require('./systemProxyEnv');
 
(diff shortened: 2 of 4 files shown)


XiaoZ-0218:
Superseded by #498, which already includes these changes — closing to keep the review in one place.

PR #490, opened 2026-08-23
feat(volcengine): track the Agent Plan alongside the Coding Plan

echohn:
## Summary

Tracks the Volcengine Ark **Agent Plan** (AFP) quota via the `GetAFPUsage` OpenAPI action, reported as a second row of the existing `volcengine` provider rather than as a separate provider.

Both plans are subscriptions on one Volcengine account: same host, same service, same signature, same AK/SK, only the OpenAPI Action differs. So `fetchVolcengineLimits` queries both and returns an array of rows. `probeLimitProvider` already flattens fetcher results, and `limits.js` dedupes on `provider:accountKey`, so the two plans survive as two rows with distinct account keys and the limits panel groups them into one card the way it already groups MiMo and Cursor accounts.

Only a Volcengine access key (signed mode) can query the OpenAPI; a plain Ark API key cannot, so it never attempts `GetAFPUsage`.

## Behaviour

- an account with only a Coding Plan is unchanged: one row, same `accountKey`, so history and burn rate keep their identity
- an account with only an Agent Plan reports just that row
- an account with both reports both, in local and in sync mode alike
- an inherited key that cannot see an Agent Plan reports nothing for it, since having no Agent Plan is the normal case
- a credential override the user actually typed in reports its failure instead, since that override is a claim that a second account exists

## What's included

- `src/shared/volcengineLimits.js` — `GetAFPUsage` fetcher and AFP window parser next to the existing Coding Plan path, reusing `signVolcengineRequest`
- `src/shared/limits.js` — Volcengine joins the multi-account list in `providerCollapseKey`, so the hub no longer collapses the two plans into one row in sync mode (hence the new Hub build revision)
- Renderer: Volcengine account group in the limits panel, plan-named row titles, and an optional collapsed credential override in the Volcengine settings card for the rare case where the two plans sit on different accounts
- Docs: README ×5, `.env.example`
- Tests: Agent Plan coverage in `tests/shared/volcengineLimits.test.js`, aggregation guard in `tests/shared/limits.test.js`, renderer guards in `tests/electron/`

## Not included

`AFPDaily` is not mapped. `normalizeWindowKind` only accepts session, weekly and billing, and `renderProviderWindows` picks those three kinds explicitly, so a daily window would be dropped twice over. Supporting it is a shared-schema plus renderer change that also touches the tray, the macOS widget and the Home overview, so it is left to its own PR.

## Verification

`npm run verify` (lint + test) passes: 3721 tests, 3720 pass, 1 skipped.

Refs #519

---
## Summary by cubic
Tracks the Volcengine Agent Plan (AFP) quota via the OpenAPI `GetAFPUsage` action, reporting it as a second `volcengine` limits row next to the Coding Plan rather than as a separate provider.

Both plans usually sit on one Volcengine account, so Agent Plan credentials fall back to the Coding Plan AK/SK; an optional, collapsed override covers the case where they were bought on different accounts. An Ark API key cannot sign OpenAPI calls, so it never queries the Agent Plan, and the settings note now leads with the access key's coverage of both plans.

| Area | Before | After |
| --- | --- | --- |
| Limits reporting | `volcengine` reports only the Coding Plan | Also reports the Agent Plan's 5-hour/weekly/monthly AFP windows, as a second row |
| Credentials | Only Coding Plan AK/SK or Ark API key | Agent AK/SK override, falling back to Coding Plan keys; Ark API key and half-filled pairs are refused for AFP |
| UI/UX | One Volcengine card with a single Coding Plan row | Both plans grouped in one card, rows titled by plan; collapsed Agent override shows a "Set" tag and a clear control once stored |
| Sync/aggregate | `aggregateLimits` collapsed Volcengine to one row per device | Both plan rows survive; distinct account keys keep them apart |

**Notes, tests, risks**
- `fetchVolcengineLimits` now returns a row array; Coding-Plan-only accounts keep their single-row output unchanged.
- An inherited key answered with 401/403/404 reports nothing for the Agent Plan; transient failures keep the row with their real status.
- An override that cannot sign OpenAPI refuses rather than silently reusing the Coding Plan key, and the card can drop a stored override without clearing the Coding Plan.
- AFPDaily is not mapped because the shared limits schema only carries session/weekly/billing.
- Unit and UI tests cover credential fallback and refusal, window parsing, error mapping, and aggregation; no migrations or breaking changes.

<sup>Written for commit 16722225d8bc32c19c529af3a0de75e5d036ff1c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/490?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


echohn:
Note on the `timezones` CI job: it currently fails, but **this failure is pre-existing on `main` and unrelated to this PR**. I reproduced the failing test (`collector publishes live periods when only Qoder CN history read fails`) on a clean `upstream/main` checkout under `TZ=America/Los_Angeles` — it fails there too (`TypeError: Cannot read properties of undefined (reading 'perClient')`). Every other check on this PR passes, and the changes here are timezone-stable (verified across all 7 offsets the job exercises).

Javis603:
Thanks @echohn, the `GetAFPUsage` mapping and the AK/SK fallback were both right.

I pushed a follow-up commit (c0c1c2de) that makes the Agent Plan a second row of the existing `volcengine` provider instead of a separate `volcagent` one. Both plans sit on one account behind the same signature, so a second provider surface was buying a row the limits panel already renders for MiMo and Cursor accounts. `fetchVolcengineLimits` returns an array now, and a Coding-Plan-only account keeps its exact `accountKey`, so existing history and burn rate are untouched. The separate-account case survives as a collapsed override in the Volcengine card.

`AFPDaily` is still unmapped: `normalizeWindowKind` only accepts session, weekly and billing, so a daily window would be dropped downstream. I will take that separately.

PR #498, opened 2026-08-24
feat(grok): manage the Grok OAuth token in the widget's own credential store

XiaoZ-0218:
## Problem

On macOS with a system-level proxy, the Grok limit card repeatedly degrades to `unavailable` → `unauthorized`. Root cause: the widget depends on the Grok CLI's `~/.grok/auth.json`, whose access tokens expire after **6 hours** and only refresh when the CLI itself runs — and the CLI is a plain Node program that needs proxy env vars a GUI-launched widget can never give its child processes. #485 fixes the child-env half; this PR removes the dependency on the CLI entirely.

## Approach

Let the widget manage its own Grok credential, exactly like opencode/kimi/deepseek credentials are managed today:

- **`refreshGrokAccessToken()`** in `grokLimits.js`: exchanges a stored refresh_token for a fresh access_token at `auth.x.ai/oauth2/token` (standard OAuth2 refresh grant, published in xAI's OIDC discovery) through the widget's existing system-proxy-aware fetch.
- **Managed path in `fetchGrokLimits`**: when `grokRefreshToken` + `grokClientId` are configured (credential store or `GROK_REFRESH_TOKEN`/`GROK_CLIENT_ID` env), the widget refreshes its own token and calls the billing endpoint directly — no CLI spawn, no `auth.json` read. On failure it **falls through to the existing auth.json auto-detection**, so nothing regresses.
- **Credential wiring**: `grokRefreshToken`/\`grokClientId` added to `CREDENTIAL_SETTING_PATHS` and `limitsConfigFromSettings`, so they persist in `credentials.json` like every other provider credential.

## Why this is safe with the CLI

xAI uses **soft refresh-token rotation** — verified against auth.x.ai: an old refresh_token remains valid after a new one is issued. So the CLI and the widget can each hold their own copy and refresh independently without invalidating each other. The widget never writes to `~/.grok/auth.json`.

## Verification

- Live-verified against xAI: refresh_token → new access_token (6h) → billing endpoint returns valid quota data (grpc-status 0). Old refresh_token still works after rotation.
- New tests: 10 cases (exchange success, rotation semantics, 401→unauthorized / 500→unavailable mapping, missing-token rejection, managed-path priority, fallback to auth.json on refresh failure, env fallback, no-refresh-token opt-out).
- Full suite: **3630 passing**, eslint clean.

## Relationship to #485

This branch builds on the child-env system-proxy fix (#485, currently open). If #485 lands first I'll rebase; if you'd prefer them merged together, the delta here is isolated to grokLimits.js + credential wiring + tests.

---
## Summary by cubic
Manages the Grok OAuth token in the widget and injects the OS proxy for Grok CLI spawns to stop unauthorized/unavailable flapping on proxied macOS networks. Previously the widget depended on the CLI’s ~/.grok/auth.json refreshing only when the CLI ran; now the widget refreshes its own access token and still falls back to the CLI path when needed.

- Adds a managed OAuth2 refresh flow (refresh token → access token) via xAI’s OIDC token endpoint.
- Prefers a managed path in fetchGrokLimits when grokRefreshToken + grokClientId are configured; otherwise auto-detects ~/.grok/auth.json as before.
- Persists grokRefreshToken and grokClientId in the credential store; also reads GROK_REFRESH_TOKEN and GROK_CLIENT_ID from env.
- Injects the macOS system proxy into child process env for the Grok agent RPC; primes proxy lookup at startup; opt-out supported.

| Area | Before | After |
| --- | --- | --- |
| Token management | Access tokens read from CLI’s ~/.grok/auth.json; refresh only when CLI runs | Widget exchanges refresh_token for access_token directly via OAuth2; independent of CLI |
| Proxy handling for CLI | GUI-launched child processes lacked proxy env; token refresh and RPC often failed on proxied macOS | Child env inherits primed system proxy (HTTPS_PROXY/HTTP_PROXY/ALL_PROXY) unless explicitly set or opted out |
| Configuration | No first-class Grok credentials in widget | grokRefreshToken and grokClientId stored in credentials.json; env fallbacks GROK_REFRESH_TOKEN and GROK_CLIENT_ID |
| Fallback behavior | Single path via CLI/auth.json | Managed path preferred; on refresh failure, falls back to auth.json/CLI without regression |

**Notes**
- Safe concurrency: xAI uses soft refresh-token rotation; widget and CLI refresh independently without invalidating each other.
- Tests: Added coverage for refresh exchange and error mapping, managed-path priority and fallback, env-based configuration, and system proxy injection; stabilized a Qoder CN history guard fixture to use the local calendar day.
- Rollout/migration: To enable the managed path, provide grokRefreshToken and grokClientId in settings or via env. No breaking changes; existing CLI-based behavior continues to work.

<sup>Written for commit 8e8cf935c66e807e49742de1c918aea4c674a9bc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/498?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/electron/runtimeConfig.js
+++ b/src/electron/runtimeConfig.js
@@ -191,6 +191,8 @@ function limitsConfigFromSettings(settings = {}, context = {}) {
     kimiApiKey: settings.kimiApiKey || '',
     kimiWebAccessToken: settings.kimiWebAccessToken || '',
     ollamaCookie: settings.ollamaCookie || '',
+    grokRefreshToken: settings.grokRefreshToken || '',
+    grokClientId: settings.grokClientId || '',
     codexManagedAccounts: context.codexManagedAccounts ?? settings.codexManagedAccounts ?? [],
     mimoManagedAccounts: context.mimoManagedAccounts ?? settings.mimoManagedAccounts ?? [],
     thirdPartyProfiles: settings.thirdPartyProfiles || {}
--- a/src/shared/credentialStore.js
+++ b/src/shared/credentialStore.js
@@ -31,7 +31,9 @@ const CREDENTIAL_SETTING_PATHS = Object.freeze({
   kimiApiKey: ['providers', 'kimi', 'apiKey'],
   kimiWebAccessToken: ['providers', 'kimi', 'webAccessToken'],
   ollamaCookie: ['providers', 'ollama', 'cookie'],
-  thirdPartyProfiles: ['providers', 'thirdparty', 'profiles']
+  thirdPartyProfiles: ['providers', 'thirdparty', 'profiles'],
+  grokRefreshToken: ['providers', 'grok', 'refreshToken'],
+  grokClientId: ['providers', 'grok', 'clientId']
 });
 
 function emptyDocument() {
--- a/src/shared/grokLimits.js
+++ b/src/shared/grokLimits.js
@@ -17,12 +17,14 @@ const { spawn } = require('node:child_process');
 const { normalizeLimitProvider } = require('./limits');
 const { hashKey } = require('./hashKey');
 const { createOutboundFetch } = require('./outboundFetch');
+const { withSystemProxyEnv } = require('./systemProxyEnv');
 const { abortError } = require('./probeDeadline');
 
 const GROK_WEB_BILLING_GRPC_URL = 'https://grok.com/grok_api_v2.GrokBuildBilling/GetGrokCreditsConfig';
 const GROK_KEY_NAMES = ['GROK_BEARER_TOKEN'];
 const GROK_OIDC_PREFIX = 'https://auth.x.ai::';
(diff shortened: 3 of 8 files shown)


XiaoZ-0218:
Friendly ping — CI is green and this merges cleanly. Note this PR also includes the system-proxy injection from #485 (now closed as superseded), so merging this one covers both. Would appreciate a review when you have time 🙏

PR #462, opened 2026-08-20
feat(qodercn): add Qwen3.8-Max model mapping and discover global Qoder directory

Aurara716:
### Summary

This PR enhances Qoder / Qoder CN token usage tracking:

1. **Model Name Mapping**:
   - Added \qmodel_38max\ -> \Qwen3.8-Max\ mapping to \QODER_CN_MODEL_DISPLAY_NAMES\.
2. **Multi-Source Discovery**:
   - Extended \qoderCnDataPaths\ to automatically discover both \QoderCN\ and global \Qoder\ data directories (\SharedClientCache/cache/db/local.db\) across macOS, Windows, and Linux.
3. **Tests**:
   - Updated \	ests/shared/qoderCnUsage.test.js\ to assert the new model code and platform path resolutions. All 20 unit tests pass cleanly with zero regressions.

### Verification
- \
ode --test tests/shared/qoderCnUsage.test.js\ (20/20 tests passed).

---
## Summary by cubic
Adds display mappings for Qwen3.8-Max, Qwen3.7-Flash, and Kimi-K2.7-Code; discovers the global Qoder data directory alongside QoderCN; and introduces built-in default pricing so costs remain visible when remote lookups fail.

- Map model codes: qmodel_38max → "Qwen3.8-Max", q37fmodel → "Qwen3.7-Flash", kmodel_latest → "Kimi-K2.7-Code".
- Discover both QoderCN/.../local.db and Qoder/.../local.db across macOS/Windows/Linux; `TOKEN_MONITOR_QODER_CN_DB_PATH` still overrides.
- Prefer remote pricing lookup when available; fall back to built-in defaults for official models if unavailable.

| Area | Before | After |
| --- | --- | --- |
| Pricing when lookup unavailable | Cost unavailable | Built-in default pricing applied for official models |
| Model code qmodel_38max | Not mapped; displayed as raw code | Mapped to "Qwen3.8-Max" |
| Model code q37fmodel | Not mapped; displayed as raw code | Mapped to "Qwen3.7-Flash" |
| Model code kmodel_latest | Not mapped; displayed as raw code | Mapped to "Kimi-K2.7-Code" |
| Default DB discovery | Only QoderCN/.../local.db probed | QoderCN/.../local.db and Qoder/.../local.db probed across macOS/Windows/Linux |

**Validation and risks**
- All unit tests pass (20/20).
- No breaking API changes; message-level dedup prevents double counting if both DBs exist; unmapped/custom models remain cost-unavailable.

[...]

PR #301, opened 2026-07-31
feat(collector): add Qoder CN local usage adapter

REMIXPLAYER:
## Summary

Adds opt-in local token-usage collection for Qoder CN, the Chinese build of Qoder, as the Qoder CN half of #164. Qoder Work CN remains out of scope because its official channel currently exposes zero token fields.

The implementation follows the existing local-adapter pattern and keeps Qoder-specific SQLite parsing, model mapping, pricing boundaries, and reconciliation isolated behind the new `qodercn` client ID.

## What changed

- Read Qoder CN's local `chat_message` SQLite data through the existing CLI/Node SQLite fallback.
- Resolve per-platform database paths and support `TOKEN_MONITOR_QODER_CN_DB_PATH` overrides.
- Normalize cached input/output tokens, deduplicate messages, and map Qoder model codes to display names.
- Keep Qoder `Auto` usage visible but unpriced because the local record does not identify the selected model.
- Keep Qoder CN model mapping, pricing lookup normalization, six-hour caching, and row-cost calculation inside the Qoder CN adapter. It uses the existing low-level `tokscale pricing` command through the collector boundary; Qoder CN does not call or import Proma business functions. Unknown/offline models remain cost-unavailable instead of guessing prices.
- Add opt-in settings, renderer, tray, Discord label integration, icon, README, collector, and watcher integration. The Discord Rich Presence asset remains pending maintainer registration.
- Keep Qoder database sidecar events from triggering a self-watch loop while continuing to watch real database changes.
- Normalize `gmt_create` in anchored reads across storage formats — numeric seconds/milliseconds, ISO text, and numeric text strings — so second- and millisecond-based Qoder databases are handled consistently.
- Include the resolved Qoder database path in the persisted anchor fingerprint so a path override cannot reuse a snapshot from another database after a transient read failure.
- Attribute sessions to projects via `chat_session.project_name` (scalar subquery on the indexed `session_id`, ~8 ms on the real database): Qoder CN usage appears in the Projects view alongside the tokscale clients, sharing the same label normalisation. Databases without the `chat_session` table fall back to the plain queries via a per-database-path probe (only positive results are cached), and the `.` "no project" sentinel stays unattributed.

## Failure isolation

- A transient Qoder read failure (period or anchored/watch, with or without a fallback) keeps the last complete Qoder period, publishes the fresh data collected for every other client, and keeps the failed reconciliation scheduled for retry.
[...]

PR #510, opened 2026-08-26
feat(qodercn): resolve model display names dynamically and inherit sub-agent models

shuming9315-Fuxiao:
## Summary

The Qoder CN local adapter already attributes each usage row to a model via `chat_message.model_info.model_key`, but two gaps make the attribution wrong in practice:

1. **The static display-name table goes stale.** On a current install, `gmodel` renders as "GLM-5" while the app itself reports GLM-5.3, and brand-new codes (e.g. `qmodel_38max`) pass through as bare codes. Every Qoder model refresh currently needs a mapping PR (see #462 adding three more codes).
2. **Sub-agent sessions have no model at all.** `session_type='agent_sub_computeruse'` sessions (the local computer-use sub-agents) record an empty `model_info` on every message — on a real database that was 459 of 697 assistant rows collapsing into the `qoder-agent` bucket.

This PR fixes both at the source:

- **Dynamic names:** the adapter now reads the model catalogue Qoder itself maintains in `state.vscdb` (`aicoding.modelConfigs.cache.*`, `{name, displayName}` entries) and prefers it over the static table. The static `QODER_CN_MODEL_DISPLAY_NAMES` stays as a fallback for when the catalogue can't be read; a missing catalogue never fails the scan.
- **Sub-agent inheritance:** rows with no `model_info` inherit the parent session's model (via `chat_session.parent_session_id` → `preferred_model_info`, falling back to the parent's latest non-empty message) and render with a ` (sub-agent)` suffix. The suffix is deliberate: the sub-agent's actual model is not recorded locally, so the suffixed name deliberately fails pricing lookup and stays cost-unavailable, keeping the "no pricing for unattributed usage" boundary from #301 while still showing which parent task model drove the usage.

Measured on a real database (697 assistant rows): unknown-model rows dropped 526 → 0 (`GLM-5.3` ×171, `Qwen3.8-Max` ×67, `Qwen3.8-Max (sub-agent)` ×459).

No changes to the static mapping table itself, so this does not conflict with #462.

## Files

- `src/shared/qoderCnUsage.js` — dynamic catalogue reader (read once per scan), sub-agent inheritance, ` (sub-agent)` suffix
- `tests/shared/qoderCnUsage.test.js` — dynamic-over-static precedence, static fallback, bare-code passthrough, unreadable `state.vscdb`, sub-agent inheritance + suffix, no-parent fallback, degraded columns

## Testing

- `node --test tests/shared/qoderCnUsage.test.js` — 35/35 (was 20 before this PR)
- Day-boundary fixtures follow the local-midnight anchoring convention from #482 (no `Z`-suffixed literals)
- End-to-end against the real `local.db` + `state.vscdb`: model distribution above; catalogue read and inheritance verified on live data.

### DIFF
--- a/src/shared/qoderCnUsage.js
+++ b/src/shared/qoderCnUsage.js
@@ -44,6 +44,27 @@ const QODER_CN_READ_MAX_BYTES = 50 * 1024 * 1024;
 const QODER_CN_READ_MAX_ROWS = 100_000;
 const QODER_CN_NEGATIVE_SCHEMA_CACHE_TTL_MS = 5 * 60 * 1000;
 const QODER_CN_READ_BUDGET_ERROR = 'QODER_CN_READ_BUDGET_EXCEEDED';
+// Qoder CN also keeps a live model catalog (code -> displayName) in the
+// VS Code-style state DB next to the app root: ItemTable rows keyed
+// `aicoding.modelConfigs.cache.<surface>`. The catalog ships with the app and
+// changes between releases (gmodel has been GLM-5, GLM-5.2 and GLM-5.3), so the
+// static table above ages while rows keep referencing the current codes. The
+// dynamic catalog wins over the static table; codes absent from both pass
+// through unchanged. `secret://`-prefixed keys in the same table are DPAPI
+// encrypted and deliberately never touched — the three catalog keys are plain.
+const QODER_CN_STATE_DB_SUFFIX = path.join('User', 'globalStorage', 'state.vscdb');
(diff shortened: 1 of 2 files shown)


Javis603:
Thanks for the real-database work on dynamic model names and parent-session attribution. Current Qoder CN builds expose model identity directly in the JSONL source being handled in #753, so I am not planning to extend the legacy SQLite/state.vscdb path further.

I am closing this one rather than asking for a rebase of the old approach. The investigation here was still useful in establishing how the previous storage format represented model and sub-agent attribution.

PR #515, opened 2026-08-26
fix(trae): handle unconsumed packs and feature entitlements in the credits response

shuming9315-Fuxiao:
## Summary

The Trae CN credits parser shipped in v0.48.0 (#483) fails closed on two pack shapes that real CN accounts return, so the provider shows unavailable instead of a balance:

- **Untouched packs report an empty `usage` object** — a welfare/invite/gift pack nobody has drawn from yet has no `credits_amount` at all. `packUsed === null` currently throws `unusable active entitlement pack`. An untouched pack is zero consumed, not malformed.
- **Free-tier feature entitlements carry a `quota` without `credits_limit`** — e.g. the 免费 pack whose quota only holds `enable_solo_*` toggles. `packLimit === null` currently throws. Those are feature switches, not spendable balance, and should be skipped.

Both shapes appear in the very first response of a CN account holding welfare/invite packs, so the provider never recovers for those users.

## What changed

`parseTraeEntUsage` in `src/shared/traeLimits.js`:

- A pack whose quota has no `credits_limit` and no usage is a feature entitlement and is skipped. If such a pack — or a pack with no quota at all — reports spend, the parser still fails closed: a shape whose spend cannot be summed stays unsafe to publish.
- A pack with a `credits_limit` but an empty `usage` counts as 0 consumed.
- Zero-limit skip, negative-value and overage handling, and every existing fail-closed error case are unchanged.

## Verification

- Live CN account carrying both shapes (5 packs: 2 untouched welfare packs with `usage: {}`, 1 feature entitlement with `enable_solo_*` only, 1 monthly pack mixing `credits_limit` + toggles, 1 partially consumed pack). v0.48.0 throws; with this change the aggregate matches the API's own `usage_summary` exactly — 5000 limit / 777.54 consumed / 4 active packs.
- New tests: real-world shape aggregation, and fail-closed when a feature entitlement reports spend.
- `node --test tests/shared/traeLimits.test.js` — 11/11 (9 existing + 2 new)
- `npx eslint src/shared/traeLimits.js tests/shared/traeLimits.test.js` — clean

### DIFF
--- a/src/shared/traeLimits.js
+++ b/src/shared/traeLimits.js
@@ -74,18 +74,30 @@ function parseTraeEntUsage(body) {
     const rawLimit = pack?.entitlement_base_info?.quota?.credits_limit
       ?? pack?.entitlementBaseInfo?.quota?.creditsLimit;
     const rawUsed = pack?.usage?.credits_amount ?? pack?.usage?.creditsAmount;
+    const quota = pack?.entitlement_base_info?.quota ?? pack?.entitlementBaseInfo?.quota;
     const packLimit = numberOrNull(rawLimit);
     const packUsed = numberOrNull(rawUsed);
 
+    // A pack whose quota carries no credits_limit is a feature entitlement
+    // (free-tier solo toggles), not balance, and contributes nothing. Usage on
+    // such a pack — or a pack with no quota at all — is a shape whose spend we
+    // cannot sum, so a partial aggregate stays unsafe to publish.
+    if (packLimit === null) {
+      if (quota && packUsed === null) continue;
+      throw new Error('Trae credits response contains an unusable active entitlement pack');
+    }
+    // An untouched pack reports an empty usage object: zero consumed, not malformed.
+    const consumed = packUsed === null ? 0 : packUsed;
+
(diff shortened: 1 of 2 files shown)


Javis603:
Thanks @shuming9315-Fuxiao for tracing these two real Trae CN response shapes and adding focused regression coverage. Treating untouched packs as zero-consumed and excluding feature-only entitlements matches the API’s aggregate while still failing closed when spend cannot be safely aggregated.

shuming9315-Fuxiao:
Thank you for the careful review and for merging #515. I’m glad the
real-world Trae CN response samples and regression coverage were
useful. I’ll keep future parser changes narrowly scoped and fail
closed whenever usage cannot be safely aggregated.

nianqixu:
> Thank you for the careful review and for merging #515. I’m glad the real-world Trae CN response samples and regression coverage were useful. I’ll keep future parser changes narrowly scoped and fail closed whenever usage cannot be safely aggregated.

你们是怎么连接上trae的，我按照指示根本找不到对应的请求包，同样的问题已经返回到最近issus，希望能找到解决方法

PR #521, opened 2026-08-26
fix(collector): cap watch-debounce deferral with a max-wait ceiling

### LINKED ISSUE #520: Live collection mode can starve indefinitely — watch debounce re-arms with no maximum wait
## Summary

In `collectionMode: "live"`, a scan is triggered only by the file watcher. The debounce in `scheduleTick` clears and re-arms on every watch event, and has no ceiling. On a session tree that is written to more often than once per `watchDebounceMs` (1500 ms), the timer resets forever and `runTick` is never reached, so the app collects nothing at all — no usage totals, no refresh, indefinitely.

Switching to `collectionMode: "interval"` restores collection immediately.

## Environment

- Token Monitor 0.48.0
- macOS 15.6.1 (24G90), arm64
- `collectionMode: "live"`, `collectionIntervalMs: 300000`, `watchDebounceMs: 1500` (default)

## The code

`src/shared/collector.js:3863-3882` on `main`:

```js
function scheduleTick(reason, eventClients) {
  if (stopped) return;
  recordWatchClients(eventClients);
  if (debounceTimer) clearTimeout(debounceTimer);
  debounceTimer = setTimeout(() => {
    ...
  }, watchDebounceMs);
}
```

Every watch event clears the pending timer and starts a new full `watchDebounceMs` window. There is no "fire anyway after N ms" ceiling, so a steady event stream postpones the tick without bound.

## Reproduction shape

Any setup where watched session files are appended to more frequently than every 1.5 seconds will reproduce it. What triggered it here:

- A tracked client's session tree of roughly 7,800 JSONL files across roughly 2,200 directories
- Ten or more agent sessions running concurrently, each appending to its own JSONL as it works
- The process held over 15,000 open file descriptors on the watched trees, so the watcher was fully armed — this is not a watcher-failed-to-attach case

## Evidence

- Under `live`, `collector-anchor.json` stayed frozen for over 24 hours: `dateKey` and `fullScanAt` never advanced past the previous day, while the app ran continuously.
- Changing only `collectionMode` to `interval` and restarting produced a completed scan within seconds, and `session-usage-archive.json` has rewritten on a steady cadence since.
- No other change was involved. The app binary, the data, and the scanner were all verified working independently: running the bundled `tokscale` by hand returned full results for every tracked client in under 2 seconds.

## Workaround

Set collection cadence to a fixed interval instead of live.

## Suggested direction

Track the timestamp of the first deferral in a debounce run and force `runTick` once a maximum wait is exceeded, so a continuous event stream degrades to a bounded cadence rather than silence.

Worth noting the existing constraint: the comment inside the timer explains that re-arming is deliberate, to avoid queueing onto an in-flight tick and stacking scans, and references the issue #15 self-trigger loop. A ceiling would need to respect both. Happy to send a PR if you can say which way you want it bounded.

KhatibFX:
Standalone live-latency fix. **This does not fix #520** — see https://github.com/Javis603/token-monitor/issues/520#issuecomment-5490829279 for the root cause of that issue, which is `child_process.spawn` failing with `EBADF` once the watcher pushes the process past roughly 10,220 open file descriptors. Bounding when `runTick` fires cannot help a spawn that fails instantly, so I have dropped the `Fixes #520` line.

What remains here is a real but narrower problem, which @Javis603 described as worth fixing on its own.

## The defect

Every watch event clears the pending debounce timer and re-arms a fresh `watchDebounceMs` window, and nothing bounds how many times that can happen. A source that changes more often than once per `watchDebounceMs` therefore defers the watch-triggered refresh without limit. The independent interval path still runs, so this is a live-refresh latency bug, not a total stall.

## This upholds the no-cooldown design, it does not reverse it

`AGENTS.md` says there is deliberately **no cooldown**, because the product promises 3–5 s updates and a mid-tick watch event must re-arm the debounce rather than coalesce. That decision stands and the re-arm is untouched. The only thing capped is *unbounded* deferral. The default ceiling is 5000 ms — the far end of the 3–5 s promise — so it can only fire in a state where the promise was already broken. In every case where the current behaviour meets the promise, the ceiling never comes due and nothing changes.

## The change

`src/shared/collector.js`:

- `WATCH_MAX_WAIT_MS` constant, 5000.
- A `watchMaxWaitMs` option, floored at `watchDebounceMs`. A ceiling shorter than one debounce window would pull every lone event forward and make the debounce itself unobservable.
- `watchDeadlineAt`, pinned by the first event of a run and 0 when no run is pending.
- Each re-arm now waits `Math.max(1, Math.min(watchDebounceMs, watchDeadlineAt - nowMs))`. The 1 ms floor is there because an expired deadline must arm a real timer rather than spin.
- The deadline is cleared on both exits from the timer callback: the `tickInFlight` re-arm and the real fire.

`AGENTS.md` point 4 is updated to describe the ceiling alongside the no-cooldown rule it qualifies.

## What is unchanged

- The `tickInFlight` re-arm — the guard against the issue #15 self-trigger loop — is unchanged. A mid-tick event still re-arms instead of queueing onto the in-flight tick.
- The forced tick goes through the same `runTick(... takeWatchClients() ...)` call as before, so clients accumulated during the storm are never dropped.

## Tests

Two new cases in `tests/shared/collectorLoadGuards.test.js`:

1. A watch-event storm faster than the debounce still ticks within the max-wait ceiling.
2. A single isolated event still waits out the full debounce and does not fire early.

`npm test`: 3680 passing, 0 failing, 1 pre-existing skip. Lint clean.

## Notes for review

1. The storm test is wall-clock sensitive. It drives events every 5 ms against a 50 ms ceiling and allows 2000 ms for the tick, which is 40x the ceiling, but it is still a timing test and could flake on a slow CI runner.
2. On the `tickInFlight` re-arm the deadline is cleared and then re-pinned by the next `scheduleTick`. So a storm that lands entirely inside one long in-flight tick gets a fresh full ceiling from that point rather than the remainder of the previous one. The alternative — leaving the deadline expired — fires that branch every millisecond for the tick's whole duration, which is why it is written this way. Please confirm that reading is what you want.
3. This branch has only ever run in unit tests. It was never executed against a real session tree, so it has no runtime evidence behind it.

Let me know if you have questions or concerns.


Javis603:
Thanks @KhatibFX for digging into the continuous-event case and for preserving the in-flight re-arm behavior.

The max-wait idea makes sense as a bound on live refresh latency, but I don't think the current PR establishes that it fixes #520.

In v0.48.0, live mode is not watcher-only. It still has an independent interval fallback: the collector runs at startup and then every `collectionIntervalMs`, which is 5 minutes in the reported configuration. `scheduleTick()` only clears and re-arms `debounceTimer`; it does not affect the interval timer.

The interval loop also forces a full reconciliation at least once per hour. The 5-minute path should therefore prevent collection from going completely silent, while a successful hourly full scan should eventually update and persist `fullScanAt`.

So a continuous watch stream can indefinitely postpone a watch-triggered refresh, but debounce starvation by itself cannot explain `collector-anchor.json` and `fullScanAt` remaining unchanged for more than 24 hours. If `tickInFlight` or another runtime path is stuck, this patch also continues re-arming and does not release that state.

I think the debounce starvation is a real live-latency issue worth fixing, but the full collector freeze reported in #520 still needs a separate explanation.

PR #524, opened 2026-08-26
feat(doubao): add Doubao Work limits provider via desktop app CDP port

shuming9315-Fuxiao:
## Summary

Doubao Work (豆包工作) is ByteDance's agent-style desktop app for productivity work. It is credit-based like Claude Code / Codex, but Token Monitor has no limits provider for it. This adds a `doubao` provider that reads the live 5-hour and weekly credit windows straight from the signed-in Doubao Work desktop app — no credentials to paste, mirroring how the WorkBuddy provider detects its desktop app.

## How it works

- The Doubao Work desktop app ships with a built-in CDP debug endpoint on `127.0.0.1:9333` (auto-detected; overridable via `TOKEN_MONITOR_DOUBAO_CDP_PORT` / `DOUBAO_CDP_PORT` or Settings).
- `src/shared/doubaoLimits.js` lists the CDP targets, reuses an existing `doubao.com` page context (attaching a hidden frame when none is open), and evaluates the same call the app's own quota page makes — `POST /alice/commerce/sale/subscription/overview/?aid=497858&web_platform=work_desktop` with cookie auth.
- The response yields the subscription plan/status plus two quota windows, mapped to the standard limits shapes: `windowType 1` → 5-hour session window (credits), `windowType 2` → weekly window. Nothing is stored beyond the normalized limits snapshot, and no new npm dependencies are introduced (the module includes a minimal RFC 6455 WebSocket client for the CDP transport).
- Both `Runtime.evaluate` expressions are plain string literals — the evaluated code is never assembled from interpolated values.

## What changed

- `src/shared/doubaoLimits.js` + `tests/shared/doubaoLimits.test.js` — provider core and tests.
- Provider registration: `limitProviders.js` (+ generated worker copy), `limitCollector.js`, hub build registry (new `core` revision via `npm run update:hub-build`).
- Settings: `doubaoCdpPort` normalization in `main.js`, runtime config, `.env.example`.
- UI: provider order/presentation, i18n connection strings (en/zh-CN/zh-TW/ko/ja), macOS widget label fallback, `styles.css` icon mask (reuses the existing `assets/icons/doubao.svg`), tool icon, README ×5.

## Verification

- Live standard-plan account: 5-hour window 9/735 credits, weekly 9/2100, subscription `active` — matches the app's own quota page exactly, including reset timestamps.
- `node --test` on the doubao suite plus the provider/README/cursor suites — 521 pass / 0 fail.
- `npx eslint` on all changed JS files — clean.
- Hub build registry regenerated; recomputed build ids match the committed registry for all three components.


shuming9315-Fuxiao:
Rebased onto main (after #523) and addressed the CodeQL finding: both `Runtime.evaluate` expressions in `doubaoLimits.js` are now plain string literals with no interpolated values, since `Runtime.evaluate` compiles the string as code. Nothing user-controlled ever reached those strings before (the interpolated constants were module-level literals), but the literal form removes the code-construction pattern entirely, so the alert should clear on the next scan.

Javis603:
Thanks @shuming9315-Fuxiao for the original implementation and live-test details.

I tested the current PR head (`19ab141b`) on Windows 11 with Doubao Work `147.0.7727.149` under a normal launch. The app was signed in, and its own quota page displayed the Standard Plan with the current and 7-day limits.

However, the CDP path could not be reproduced:

- `127.0.0.1:9333` was not listening.
- The Doubao-owned port `49853` returned HTTP 404 for `/json`, `/json/version`, and `/json/list`.
- The other Doubao-owned port, `59889`, closed the connection without returning an HTTP response for those endpoints.
- None of the running Doubao processes had `--remote-debugging-port` or `--remote-debugging-pipe` enabled.

The provider therefore remained `notConfigured`. This also shows that discovering Doubao's dynamic listening ports would not be enough, since neither port exposes a standard CDP endpoint.

Together with the macOS investigation in #586, the standard CDP path does not appear to be available in current clients under a normal launch. Could you share the exact OS and Doubao Work version/build used for your successful test, along with any launch flags or configuration steps that were required?

Until the endpoint can be reproduced safely on a current client, I don't think this implementation is ready to merge.

shuming9315-Fuxiao:
Thanks for the detailed reproduction. I went back to the original August 26 investigation directory and recovered the network-response captures and CDP scripts, in addition to the unit tests.

The archived request User-Agent identifies **Doubao Work 2.25.18** (`SamanthaDoubaoWork/2.25.18`) on Windows x64 with Chromium 147. The saved response from `/alice/commerce/sale/subscription/overview/` has `code: 0`, Standard Plan, and these fields:
- `window_type: 1`: `used_amount: "9"`, `total_amount: "735"`
- `window_type: 2`: `used_amount: "9"`, `total_amount: "2,100"`

The original scripts request `http://127.0.0.1:9333/json`, attach to the returned Doubao page/iframe target, and capture responses with `Network.getResponseBody`. They attach to an already-running instance; they do not launch the application. I have not recovered that instance's launch arguments or the exact Windows build, so I cannot establish whether debugging was enabled through an additional step.

**The PR description's claim that the app ships with a default CDP endpoint was too broad.** The evidence supports the earlier working instance and captured quota data, but does not establish that ordinary launches expose port 9333.

I also reproduced your current-client result: Windows 11 Home 26200.9278, installed launcher 2.27.8 / application executable 147.0.7727.149, normal launch without added debugging flags. Port 9333 was not listening; the application-owned port 49853 returned 404 for both standard endpoints, and its other listener did not return an HTTP response. I did not verify the signed-in quota page in this latest check.

The version difference is a useful lead, not proof that an update caused the change. I agree the current implementation is not ready to merge without a reproducible setup on a current client. Any required debugging configuration should be documented explicitly, rather than treated as automatic discovery.

PR #453, opened 2026-08-18
feat(kimi): collect Kimi Work usage and attribute sessions to projects

REMIXPLAYER:
## Summary

Integrates Kimi Work (the Kimi desktop app's agent mode) into the existing `kimi` client:

- Mirror the Kimi Work source roots used by the pinned Tokscale build so source health and live watcher attribution cover the desktop app on macOS and Windows.
- Resolve Kimi Code and Kimi Work project identity from each session's sibling `state.json` (`workDir` / `custom.workspacePath`), while honoring the Projects opt-out.
- Classify the Kimi Work model IDs (`k2d6-agent`, `k3-agent`, `k3-agent-swarm`) as Kimi so model rows use the existing Kimi icon and colour.
- Document Kimi CLI, Kimi Code, and Kimi Work support across all localized READMEs.

Kimi Work remains part of the existing `kimi` product surface. This does not add a separate client or setting.

**Integration boundary:** usage parsing and pricing remain delegated to the pinned Tokscale build. This PR does not add a second Kimi Work parser/pricer or inject extra Tokscale scan roots; Token Monitor adds the metadata, source-health, watcher, project-attribution, and presentation integration around Tokscale's native support.

## Relationship to Tokscale

The pinned Tokscale build is the source of Kimi Work discovery, parsing, and pricing. Token Monitor mirrors its macOS and Windows roots for source health and live watcher attribution, then enriches the resulting session rows from local `state.json` metadata.

See [Tokscale PR #1170](https://github.com/junhoyeo/tokscale/pull/1170) and [merge commit b061f128](https://github.com/junhoyeo/tokscale/commit/b061f12810dc6286a2ab5477aea508a8d0f3fb10).

Closes #349.

## Why

Kimi Work uses the same embedded `kimi-code` runtime and `wire.jsonl` protocol as Kimi Code, but stores its sessions under the Kimi desktop app data directory.

The Tokscale build currently pinned by Token Monitor discovers, parses, and prices these sessions natively. Token Monitor still needs to mirror those source roots for health detection and live refresh, and enrich the resulting session rows with project metadata because Tokscale's `client,session,model` output does not expose Kimi workspace identity.

## Implementation

- Mirror Tokscale's Kimi Work roots:
  - macOS: the Kimi desktop app data directory;
  - Windows: always include the home-relative root; add the environment-derived root only when `APPDATA` is a non-empty string, using relocated `shareDir` values from `daimon-storage.json` when available;
[...]

PR #525, opened 2026-08-27
feat(kimi): desktop-app session with self-refresh and manual token pairs

### LINKED ISSUE #349: 希望能纳入kimi客户端的（kimi work）的token统计
实现感谢开发大佬（们），这是个非常好用的工具 👍

我平时也在使用 Kimi（月之暗面的 AI 工具），很希望 Token Monitor 也能把 Kimi 的 token 用量统计进来，方便和其它工具放在一起看。

（我本地使用kimi 客户端，未发现kimi被纳入了统计）

我自己不是开发，但用 AI 助手（WorkBuddy，模型 HY3）调研了一下 Kimi 在本机保存 token 用量的日志位置，并在自己电脑上做了一次本地读取验证——初步看是可行的。

不过我没有开发背景，不确定采集逻辑是否完全准确，
所以把这次调研整理成了一份技术文档（已脱敏，不含任何个人隐私信息），作为附件上传，供开发者参考实现。

希望能早日支持 Kimi，谢谢！🙏

[TokenMonitor支持Kimi接入-技术文档.md](https://github.com/user-attachments/files/30823555/TokenMonitor.Kimi.-.md)

sisibeloved:
## Behaviour change

Kimi web limits were effectively broken: the setup copy pointed at the legacy `kimi-auth` cookie, but the Kimi website no longer refreshes it (30-day JWT, never renewed — it is absent from the current frontend bundle entirely), and the localStorage `access_token` it was replaced by lives only ~15 minutes. Users pasted a fresh token and watched it die within minutes; the provider flipped to `Sign in again` with nothing they could do from the settings page.

This adds the durable source as a zero-setup default and fixes the manual path:

- **Kimi Work desktop app session (Windows, default on).** The app persists a rotating access/refresh pair to `%APPDATA%\kimi-desktop\bridge-store\token-store.json`, safeStorage-encrypted (`v10` AES-GCM under the DPAPI-protected key from its `Local State`). Token Monitor reads that pair on every limits tick, so the provider works with no manual setup at all. `TOKEN_MONITOR_KIMI_DESKTOP_SESSION=0` opts out.
- **Refresh ownership is split by app state** — this is the load-bearing constraint. The app's TokenStore loads the file once at startup (no file watcher), rotates the refresh token on every use, and escalates any 401/403 to a session-wiping `clear()`, so blindly refreshing would race the app and could log the user out. While `kimi-desktop\Kimi.exe` is running, Token Monitor stays strictly read-only (an idle-running app past the 15-minute access-token life reports `unavailable` and self-heals on the app's next use); while the app is closed, Token Monitor refreshes the pair itself via the app's own `auth.kimi.com/api/account.gateway.v1.AuthService/RefreshToken` endpoint and writes the rotated pair back in the app's exact format, which the app's next start picks up transparently. A rejected refresh token never clears the store here — the app's next real sign-in self-heals.
- **Manual refresh-token sessions.** The single paste box now classifies what it receives (JWT lifetime: minutes = access token, days = refresh token) and stores a pasted refresh token — `KIMI_REFRESH_TOKEN` for headless — as the seed of a self-renewing session whose rotated pair persists in `sharedDataDir/kimi-manual-session.json`. This finally gives non-Windows users a working manual credential. A pasted refresh token shadows the desktop session entirely, since one refresh token has exactly one consumer.
- **Setup copy** in all five locales now points at Local Storage (`access_token` / `refresh_token`) instead of the dead cookie, and the Kimi account row can show a "Kimi desktop app" source.

Two robustness fixes came out of live testing: the mtime-keyed session cache now recomputes staleness per read (a cached `accessIsStale = false` previously let an expired token slide through undetected — the provider died exactly 15 minutes after startup), and both the refresh and quota requests retry once on fast transient failures with `[kimi]`-prefixed logging on the refresh path, which was previously fully silent.

## Verification

- `npm run lint` clean; `tests/shared/kimiDesktopSession.test.js` (21) + `tests/shared/kimiLimits.test.js` (38) + the limits chain (`limitCollector` / `deviceRuntime` / `limitsRuntime` / `credentialStore`) + `tests/electron/i18n.test.js` all green (145 passed, 0 failed).
- Live end-to-end on Windows against a real Kimi Work install: idle app → `unavailable` with no doomed round trip; app closed with an expired token → self-refresh succeeds, GetUsages returns all three windows (5-hour / weekly / billing), and restarting the Kimi app afterwards proves the written-back pair is picked up losslessly (no logout). A real pasted refresh token was exercised through the manual path the same way.
- Stale-cache regression test: the same file read 30 minutes later must report `accessIsStale = true`.

Closes #349's limits half (usage attribution for Kimi Work landed separately in #453).

[\[ Run npm run verify \]]()

---
## Summary by cubic
Kimi limits now run on durable, rotating sessions from the Kimi Work desktop app on Windows or a pasted refresh token, instead of the dead `kimi-auth` cookie or short-lived access-token pastes. This keeps the 5-hour, weekly, and billing windows current with less manual reauthentication.

| Area | Before | After |
| --- | --- | --- |
| Windows session | No reliable automatic credential | Reads the encrypted desktop-app token pair and refreshes it while `Kimi.exe` is closed; the app remains the refresh owner while running |
| Manual session | Pasted access tokens expired after about 15 minutes | Pasted refresh tokens persist rotated credentials for about 90 days; pasting an access token no longer clears a stored refresh seed |
| Credential precedence | No coordinated source priority | Manual credentials shadow the desktop session; with a Code API key configured, Code stays primary while the session still supplies the windows Code doesn't return |
| Setup | Instructions referenced the inactive `kimi-auth` cookie | Instructions use Local Storage `access_token` or `refresh_token` and explain each token's lifetime |
| Initial status | Cold starts could remain stuck on "Checking" | Shows that the first usage baseline is being collected |
| Web console coexistence | Rotating credentials were shared without warning | The same account's web console signs out on the next refresh once Token Monitor owns the token pair |

**Design and compatibility**

- `TOKEN_MONITOR_KIMI_DESKTOP_SESSION=0` disables automatic desktop-session use on Windows.
- `KIMI_REFRESH_TOKEN` supports headless use; a configured `kimiWebRefreshToken` option outranks it, and a refresh-token-only session counts as configured.
- The web-token field names both accepted token types; a stale manual session reports unavailable rather than falling over to the desktop store.
- Reads recompute token staleness per read; transient network failures retry once, but the refresh exchange never retries since it rotates the token server-side; refresh failures log with a `[kimi]` prefix.
- Persisted writes go through a sibling temp file and atomic rename, leaving the previous contents intact on failure.
- When a Code API key is configured, account identity follows the key; otherwise it uses the session's stable user id.
- Setup copy and README updates cover all five locales; this closes the limits portion of #349, while Kimi Work usage attribution remains separate.

**Validation**

- `npm run lint`, the Kimi session and limits suites, limits-chain tests, and i18n tests pass with 145 tests; live Windows checks covered refresh, app-restart persistence, and pasted refresh tokens.

<details>
<summary>中文版本</summary>

Kimi 額度現已改用可持久輪換的工作階段：在 Windows 上自動讀取 Kimi Work 桌面應用程式的登入階段，或使用貼上的 refresh token，不再依賴已失效的 `kimi-auth` Cookie 或有效期短暫的 access token。這可在減少手動重新驗證的情況下，持續更新 5 小時、每週及帳單額度視窗。

| Area | Before | After |
| --- | --- | --- |
| Windows 工作階段 | 沒有可靠的自動憑證 | 讀取桌面應用程式的加密權杖對，並在 `Kimi.exe` 關閉時重新整理；應用程式執行期間仍由它負責重新整理 |
| 手動工作階段 | 貼上的 access token 約 15 分鐘後失效 | 貼上的 refresh token 會保存輪換後的憑證約 90 天；貼上 access token 不再清除已儲存的 refresh seed |
| 憑證優先次序 | 沒有協調的來源優先次序 | 手動憑證會優先於桌面工作階段；設有 Code API key 時，Code 維持主要來源，而工作階段仍填補 Code 未提供的額度視窗 |
| 設定 | 指示使用已停用的 `kimi-auth` Cookie | 指示使用 Local Storage 的 `access_token` 或 `refresh_token`，並說明各權杖的有效期 |
| 初始狀態 | 冷啟動時可能長時間停留在「Checking」 | 顯示正在收集首次用量基線 |
| 網頁主控台共存 | 未提示便共用輪換憑證 | Token Monitor 接管權杖對後，同一帳戶的網頁主控台會在下一次重新整理時登出 |

**設計及相容性**

- `TOKEN_MONITOR_KIMI_DESKTOP_SESSION=0` 可停用自動桌面工作階段。
- `KIMI_REFRESH_TOKEN` 支援無頭模式；已設定的 `kimiWebRefreshToken` 選項會優先於它，而僅有 refresh token 的工作階段亦會被視為已設定。
- 網頁權杖欄位會同時說明兩種接受的權杖類型；過期的手動工作階段會顯示為「不可用」，不會轉用桌面工作階段。
- 每次讀取都會重新計算權杖是否過期；暫時性的網路失敗會重試一次，但重新整理交換因已在伺服器端輪換權杖而不會重試；重新整理失敗會以 `[kimi]` 前綴記錄。
- 持久化寫入會經由同目錄暫存檔及原子重新命名完成，失敗時會保留原有內容。
- 設有 Code API key 時，帳戶識別會以該 key 為準；否則使用工作階段的穩定使用者 ID。
- 設定指示及 README 已更新全部五種語系；這完成 #349 的額度部分，而 Kimi Work 用量歸因仍然分開處理。

**驗證**

- `npm run lint`、Kimi 工作階段及額度測試、額度鏈測試和 i18n 測試均通過，共 145 項測試；Windows 實機檢查涵蓋重新整理、應用程式重新啟動後的持久性，以及貼上 refresh token。

</details>

<sup>Written for commit e3bd1b694588d62e1b69e82b29077d3f38e45fbb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/525?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Known gap (follow-up)

The idle-running-app staleness window — while the Kimi Work app is open but unused past the 15-minute token life, the provider reports unavailable until its next use — is deliberate (refreshing under a running app risks the app's session-wiping clear()). Tracked in #529, including a candidate early-resolve design surfaced in review.

## Summary by CodeRabbit

* **New Features**
  * Added support for Kimi renewable web refresh tokens with automatic renewal and persistent sessions.
  * Windows installations can detect and use Kimi Desktop sessions automatically when no manual credential is configured.
  * Kimi usage limits now support desktop sessions and refresh-token authentication, with improved stale-session recovery.

* **Bug Fixes**
  * Improved Kimi credential status reporting and logout behavior.
  * Added clearer baseline collection and checking states when limit data is unavailable.

* **Documentation**
  * Updated setup guidance and localized documentation for Kimi access and refresh tokens.


takenroot:
Tested this PR end-to-end on **Windows 11**, widget in local mode, no Kimi Work desktop app installed (manual paste path only). Build: `2d13856` cherry-picked onto **v0.51.0**, packaged via `npm run dist:win` (NSIS). Verified with the project's own suite first (`npm run verify`: 3942 tests, 0 failures), then live against my real account.

The self-renewing manual session works as designed: with a refresh token in place, the widget rotated the pair through `auth.kimi.com/api/account.gateway.v1.AuthService/RefreshToken` and persisted it to `%APPDATA%\Token Monitor\kimi-manual-session.json` — observed real rotations (access token re-minted with a fresh 15-minute life, refresh token re-extended to a new 90-day window), and the Kimi row kept populating (5-hour / weekly / monthly membership windows). Other providers (MiniMax) were unaffected.

Two findings from live testing:

**1. Paste-box UX trap: `access_token` vs `refresh_token`.** The Kimi console's localStorage holds both keys side by side, and copying the wrong one is easy. A pasted *access* token passes the save probe while fresh, lands in `kimiWebAccessToken`, **clears `kimiWebRefreshToken`** — then silently dies ~15 minutes later. I hit exactly this: an expired access token ended up shadowing a healthy manual session, and the lane then ignores `kimi-manual-session.json` entirely because no seed points at it. Suggestion: when a paste classifies as a short-lived access token while a refresh session already exists, warn — or keep the existing session instead of clobbering it.

**2. Cold-start UX: "checking" state can hang for minutes.** Limits are buffered until the usage baseline completes, which on a machine with many tracked clients took ~4.5 minutes here (three serial tokscale scans + WSL scan). During that window the limits page shows "no data", and a paste stays in the "checking" state the whole time, because the pending state only clears when the provider appears in a stats payload. From the user's seat this looks exactly like "the credential didn't work" — I initially concluded the PR was broken until I watched the stats payload directly. Suggestion: clear the pending state on the save probe's own result, or show a "baseline still collecting" hint instead of a bare empty state.

Also confirming the single-consumer constraint reproduces as documented: once the widget owns the chain, the web console's localStorage pair dies on the widget's first rotation (my browser session was logged out right after the widget's rotation).

sisibeloved:
Both findings are fixed in 1e8b9ac — thanks for the careful live testing, especially for catching the paste-box trap before merge.

**1. Access-token paste clobbering a healthy session** — fixed at the root. `classifyKimiWebTokenPaste` now receives the stored `kimiWebRefreshToken` and preserves it when the paste classifies as a short-lived access token, so the seed survives the mistake. On the collector side the priority chain already defers to the manual session whenever a seed exists, so a mis-pasted access token becomes a harmless no-op instead of a time bomb: the session keeps serving, and `kimi-manual-session.json` is never orphaned. Regression test added (`fetchKimiLimits lets the seed session serve once a pasted access token exists alongside it`).

**2. Cold-start "checking" hang** — the pending pill now tells the truth. `renderExternalProviderStatus` shows `settings.common.collectingBaseline` ("Collecting the first usage baseline — limits appear once the initial scan finishes...") while a check is pending and no limits summary has landed in the stats payload yet, and falls back to "checking" once limits start flowing. Applied to every provider's pending state rather than kimi-only, since the buffering behavior is the device-level `DeviceState` rule, not a kimi one. Clearing pending on a save probe (the other suggestion) would need a new probe IPC for cookie-style providers — happy to take that in a follow-up if you prefer the pending state to actually resolve early rather than just be explained.

**Copy** — the Kimi panel intro now states what each path shows and what to expect: the windows covered (5-hour / weekly / monthly), that a signed-in Kimi Work app on Windows is picked up automatically with the token renewing even while the app is closed, that only an open-but-idle app pauses updates, and for the manual path that refresh_token renews itself while access_token lasts ~15 minutes. All five locales updated, including the static no-JS fallback in index.html.

Rebased onto current `main` (v0.51.0) — single conflict in main.js's limitCollector require (zedCookie). Lint clean; kimi/session/guard/i18n suites 156 passed, 0 failed.

Regarding the confirmed single-consumer logout on the browser side: agreed it's inherent to rotating refresh tokens. The panel copy now says the desktop app is picked up automatically on Windows, which is the path that avoids the web console entirely — a note about the web session being signed out once the widget owns the chain might still be worth a line in the README's limits section if you want it.

takenroot:
Thanks for the quick turnaround — the seed-preservation fix and the baseline copy both read exactly right. On whether the pending state should resolve early via a save-driven probe: this is a design trade-off, so here are both sides as I see them, and I'm happy with whichever call you make.

**Doing it (probe-driven early resolve)**
- Gives an immediate valid/invalid answer at paste time, closing the loop on the most anxious moment of first-run setup.
- The user never needs to learn about the baseline-buffering concept at all.
- Costs: it mainly serves the cold-start window (once a baseline exists, the scoped lane refresh lands in seconds anyway); probe semantics differ per provider and some have side effects (a kimi probe consumes a token rotation, cookie-style probes risk rate limiting, some providers lack a cheap validation endpoint); and a probe path that diverges from the collector path is the kind of duplication this codebase has been bitten by before.

**Not doing it (keep the `1e8b9ac` copy approach)**
- Small, safe, no new IPC surface, no side-effect risk; the copy already removes the "it's broken" misreading; steady state is identical.
- Costs: a wrong paste still goes unanswered until the baseline completes (minutes on heavy machines); "checking" still doesn't reflect an actual credential check — it's explained, not resolved.

One middle option, in case it tips the scales either way: a credential change already triggers a scoped lane refresh in `LimitsRuntime`, so a fresh check result already exists in the main process — only its *delivery* is buffered by `DeviceState`. Surfacing that existing result to the settings panel (e.g. a lightweight check-result event) would give early resolve for every provider without probing twice. I don't know the wire-compat cost of that path, so it may well be a wash — your call.

Either way, +1 on the README line about the web session being signed out once the widget owns the chain.

sisibeloved:
Decision on the probe question, plus the README line you +1'd.

**Probe / early-resolve: deferring, with your middle option captured as the follow-up.** The probe path's costs you listed (per-provider semantics, side effects — a kimi probe would interleave with the desktop app's token rotation — and a second request path to keep in sync) tip it out of this PR. Your event-based middle option — surfacing the already-existing scoped-lane result past the `DeviceState` buffer instead of probing twice — is the right shape, and I'd rather not grow this PR's wire surface to find out whether it's a wash. I've added it to #529 as the candidate follow-up so the insight doesn't live only in this thread. If it turns out cheap, it supersedes the collecting-baseline copy.

**README note: done** in 2aeb880 — the limits bullet in all five locales now states that Kimi web credentials are single-consumer and that using the same account's web console after the widget owns a token pair signs that browser session out on the next refresh.

Head is now rebased on current `main` (three commits: the feature, the copy/guard alignment, and the review fixes), CI green across the matrix.

Javis603:
Thanks @sisibeloved for all the work here, especially the live testing around Kimi's rotating tokens and desktop-session behavior.

Before deciding how to proceed with this PR, could you help test the current `main` implementation after #700 against your live Kimi account?

Historically, the Kimi reports we've received have consistently shown the Code API providing the 5-hour and Weekly windows, with Monthly being the missing piece. For example, #181 and the earlier Weekly fixes were based on that behavior.

#696 recently exposed a newer `/coding/v1/usages` response shape where the Monthly pool is available directly under `usages.limit_month_total` / `limit_month_code`. That support has now been merged in #700, and the Code API is the preferred quota source, so Monthly no longer needs to depend on a Web/Desktop session.

The unusual part of #696 is that the reported payload also omitted the canonical top-level `usage` Weekly window. That's the first real payload we've seen with that behavior, so I'm not sure yet whether it is account/plan-specific, part of a newer rollout, or just an occasional response shape. #697 was opened based on that observation, but I'd like another live data point before treating Weekly as something that generally requires the Web path.

Since you already have a working Kimi setup, could you try the current `main` build with the Code API key and let us know whether you get 5-hour, Weekly, and Monthly?

If the Code API reliably covers all three, I'd prefer to keep the simpler API-first implementation rather than take ownership of Kimi Desktop's rotating session. The desktop path here is Windows-specific and, by default, automatically discovers and decrypts Kimi Desktop's safeStorage credentials, then has to coordinate refresh ownership and write rotated credentials back into Kimi's own store. That's a fairly large cross-app credential/session surface if it is only needed as a fallback.

I don't want to rule out local integration entirely, though. If we later find a cleaner way to cover any remaining gaps across platforms — ideally read-only, without taking ownership of Kimi's rotating refresh-token lifecycle or modifying the desktop app's credential store — I'd be happy to revisit it.

Thanks again. A real-account check of the merged #700 behavior would help a lot before we make the final call on this PR and the related #697/#529 follow-ups.

sisibeloved:
Live data from the exact test you asked for — and it answers it with "not on this account", plus some plan context that may explain why.

**Setup:** installed release **v0.59.0** (includes #700), real account, Kimi Code API key only, no web token configured.

**Result:** 5-hour ✓, Weekly ✓, **Monthly ✗** — the Kimi row renders 5-hour and Weekly and nothing else (screenshot below).

**Raw `/coding/v1/usages` response captured at the same time** (top-level keys): `usage` (the canonical Weekly window — limit 100 / used 47 / remaining 53), `limits[]` (the 300-minute 5-hour entry), and `usages` containing **only** `limit_5h` and `limit_7d`. There is **no `limit_month_total` and no `limit_month_code`** in the payload at all, so the merged #700 parser has no monthly field to read — the gap is upstream data, not parsing.

**Plan context (info, not a conclusion):** Kimi membership currently comes in at least two subscription schemes (new vs old); this account is on the **old** scheme. On the old scheme, the monthly figure appears to exist only on the web membership side — `GetSubscriptionStats` → `subscriptionBalance` — which earlier live checks against this same account confirmed (a monthly window populated around 84–91% used, while the Code API never carried a monthly pool). A plausible reading is that #696's reporter — and the account @takenroot validated the manual session with — are on the **new** scheme, which is why `limit_month_total` shows up there. That is one data point against at least two others; more account samples would be needed before treating either shape as the general case, so leaving the direction to you.

v0.59.0 with Kimi API key only — 5-hour and Weekly render, Monthly absent

If it turns out old-scheme accounts never expose `limit_month_total`, API-first would leave Monthly permanently blank for them, and a web-credential path (desktop session or pasted refresh token) would remain the only route to that window — but that's exactly the kind of call that needs the sample count above, so I'm not treating #525 as the answer yet.

**Unrelated aside, since the screenshot is here:** v0.59.0 renders the GLM Coding Plan's **MCP window twice** — two identical "MCP · 78% left · Reset 18d 1h" rows stacked above Balance. Looks like a duplicate-row bug in the GLM presentation rather than anything Kimi-related; flagging in passing.

sisibeloved:
All seven findings addressed in e3bd1b6 — thanks for the thorough pass.

1. **`kimiCredentialConfigured`** — now counts `settings.kimiWebRefreshToken`, so a refresh-token-only session reads as configured and the panel collapses once the provider reports ok.
2. **aria-label / `webTokenLabel`** — the field and all five translations now say "access or refresh token" (zh-CN/zh-TW gained actual translations, they had been carrying the English string).
3. **Token-store write-back** — persists through a sibling temp file (`pid` + random suffix), chmod applied to the temp file, then an atomic rename; a failed write unlinks the temp and leaves the previous store untouched.
4. **Refresh retry removed** — agreed on the semantics: the exchange rotates the token server-side, so a lost response can't be safely replayed. The retry loop is gone; the existing cooldown on the refresh path remains the transient-failure recovery. (The read-only quota requests in `fetchJson` keep their one fast retry — those are safe to replay.)
5. **Manual-cache encryption** — partially implemented, partially declined with rationale: the atomic-write half is done as above. The encrypt-or-skip half would fork the headless story: `safeStorage` is Electron-main-only, and the same shared module backs the headless agent, so OS-backed encryption would either add a DPAPI/spawn dependency to shared code or silently disable persistence for headless users. Plaintext-with-restrictive-permissions matches the documented trade-off of the project's own credential store (AGENTS.md: "deliberately local plaintext protected by filesystem permissions"). The cache sits in the same `sharedDataDir` with `0600` applied; happy to revisit if you'd rather trade headless persistence for encryption.
6. **Precedence** — the explicit `kimiWebRefreshToken` option now outranks the `KIMI_REFRESH_TOKEN` env default.
7. **Stale manual session** — reported `unavailable` without spending the expired token on the quota API, and it no longer falls through to the desktop store: with a manual seed in play, both stores hold tokens from the same rotating chain, so silently failing over would read as an account switch. Regression tests added for the stale case and the precedence fix.

Lint clean; kimi/session suites 47 passed, guard/i18n/readme/runtime suites 319 passed, 0 failures.

PR #547, opened 2026-08-29
spike: keep the widget above the Windows taskbar (not shipping)

Javis603:
**Not for merging.** Refs #533. Opened as a record of what was tried, measured on Windows, and why it is not shipping.

### The question

#541 fixed our own half of #533: Electron was actively placing the `floating` z-order level behind `Shell_TrayWnd`, on every `SetAlwaysOnTop` and on every window activation, so the widget sank under the taskbar and stayed there even when clicked.

What remained is Windows itself. When the taskbar takes activation it raises `Shell_TrayWnd` to the top of the topmost band, and a window is never notified when another window overtakes it. This branch asks whether an app can re-claim that position well enough to be worth shipping.

### What is in here

1. `keepAboveTaskbar`, an opt-in Windows-only setting, off by default, shown only in the floating window behavior.
2. A keeper that re-asserts z-order with `moveTop()` — `SetWindowPos(HWND_TOP, SWP_NOACTIVATE)` — while, and only while, the widget is on Windows, pinned above apps, visible, not minimized, and actually overlapping the area outside the display work area.
3. A `SetWinEventHook(EVENT_SYSTEM_FOREGROUND)` subscription through koffi, so the re-assert is event-driven instead of polled. `WINEVENT_OUTOFCONTEXT | WINEVENT_SKIPOWNPROCESS`; the native callback does nothing but hand off to the next tick, and every failure path falls back to polling.
4. Debug logging behind `TOKEN_MONITOR_TASKBAR_ZORDER_DEBUG=1`, which is how the measurements below were taken.

### What was measured on Windows

Three flows, all with the widget parked over the taskbar:

| | 250ms polling | hook, 2000ms interval | hook, 250ms interval | + follow-ups to 800ms |
|---|---|---|---|---|
| widget focused → click taskbar | flickers | flickers | clean | clean |
| other app focused → click taskbar | flickers | worse | clean | clean |
| widget → other app → click taskbar | flickers | worse | flickers | flickers |

The hook works: it installs, and it delivers every foreground change with sub-millisecond latency. It genuinely fixed the flow #533 was reported on — switching apps from the taskbar without ever touching the widget.

What it cannot fix is that the shell raises the taskbar at a point of its choosing after the event that announced the switch. Re-asserting before that raise is a no-op, so the widget is covered and then restored. That is a cover-then-restore flicker by construction: polling shortens it, the hook shortens it further, and extending the follow-ups to 800ms moved which step recovers it without removing it.

### Why it is not shipping

The setting would promise "keep above the taskbar" and visibly fail to on a flow users hit, in exchange for a permanent timer and a system-wide hook with no CI coverage, in a process that runs all day. A native callback into the main process needs a clean win to justify itself, and "two flows out of three" is not one.

Nothing better was found for a floating window. PowerToys has the same limitation open since 2023 (microsoft/PowerToys#28573), Electron has electron/electron#38020 and electron/electron#35225, and Rainmeter ships this as a separate `Stay Topmost` position with its side effects documented rather than as a default.

The apps that do solve it are not floating at all: TrafficMonitor inserts a native child window into explorer's own window tree (`Shell_TrayWnd → ReBarWindow32`) and follows the parent's `WM_SIZE`, so z-order never arises. Electron cannot reparent a Chromium window into another process's tree without breaking composition, input routing and DPI, and Windows 11's XAML taskbar removed the deskband and toolbar extension points anyway.

### If this is ever revisited

The branch is self-contained: `src/electron/windowsTaskbarZOrder.js`, `src/electron/windowsForegroundHook.js`, the setting, and 30 tests covering the geometry, the lifecycle and the hook fallback. Start from the measurement table above rather than re-deriving it. The remaining idea nobody has tried here is not another timer — it is not being a floating window.

---
## Summary by cubic
Spike for #533, not for merging: tests whether re-asserting the floating widget's z-order can keep it above the Windows taskbar. Adds an opt-in Windows-only `keepAboveTaskbar` setting that calls `moveTop()` while the widget overlaps the reserved taskbar area, driven by a `SetWinEventHook` foreground hook (via koffi) with a polling fallback.

| Area | Before | After |
| --- | --- | --- |
| Floating widget vs. Windows taskbar | Widget sinks under the taskbar when it takes activation | Opt-in re-assert keeps it above in two of three measured flows; the remaining one still flickers |
| Windows-only setting | None | `keepAboveTaskbar`, default off, shown only on Windows in floating mode |

**Why it is not shipping**
- The shell raises the taskbar after the event announcing a switch, so re-asserting covers-then-restores — the hook shortens the flicker but cannot remove it.
- The setting would promise the behavior and visibly fail on a common flow, in exchange for a permanent timer and a system-wide hook with no CI coverage in a process that runs all day.

**If revisited**
- Self-contained: `src/electron/windowsTaskbarZOrder.js`, `src/electron/windowsForegroundHook.js`, the setting, and 30 tests; start from the measurement table in the diff.
- PowerToys and Electron have the same open limitation; the apps that solve it reparent a native child into explorer's window tree, which Electron cannot do.

<sup>Written for commit fb4d49b762b1c33a87df24a4098a189c9a689210. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/547?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Javis603:
Closing as a record rather than a change: the measurements are in the description, and #533 gets the explanation. Reopen only if someone finds an approach that is not another timer.

Issue #550, opened 2026-08-29
refactor: define integration boundaries before structural cleanup


Javis603:
Token Monitor has accumulated a growing number of cross-cutting integration touch points as tracked clients and limits providers have expanded. Adding a tracked client currently requires coordinated edits across roughly a dozen places: collection, source discovery and health checks, WSL handling, renderer presentation and icons, tray and Discord presentation, settings, and documentation. #496 documents this well. Limits providers have a better one-file-per-provider shape, but registration, credentials, account identity, and presentation are still spread across several layers.

The same growth has concentrated unrelated responsibilities in a few files. `src/electron/main.js` and `src/electron/renderer/app.js` have both become central integration points that most feature work has to pass through, and `src/shared/collector.js` and `src/shared/limitCollector.js` now hold provider-specific behaviour alongside orchestration. The practical effect is that unrelated changes collide in the same files and are harder to review.

I want to address this incrementally rather than start with a large file or directory reorganization.

The direction:

1. Add guard tests around the current integration invariants, so missing or inconsistent wiring fails CI instead of surfacing as a user report. These should assert relationships (for example, that every known client has the required presentation metadata, and that source-check or WSL mappings do not contain orphan client ids) rather than pin the current file layout, so they keep their value through the later steps.
2. Define clearer usage and limits integration contracts and reduce the duplicated registration points. Account and credential handling is the first boundary I plan to review, since several in-flight changes overlap there.
3. Decompose the orchestration files once those boundaries exist: `collector.js` and `limitCollector.js` toward coordinating registered integrations rather than containing them, `main.js` toward application composition, and `app.js` toward feature-owned renderer modules instead of one global controller.
4. Reorganize files and directories only after ownership boundaries are stable. Doing it earlier mostly produces merge conflicts without reducing the coupling.

Not in scope: rewriting the app, adding a framework or a build-time codegen step for this, generating translated documentation, forcing every integration through one universal interface, or combining usage and limits into a single runtime lifecycle. Public behaviour, the wire format in `docs/API.md`, and stored settings should remain compatible by default unless a focused step explicitly requires and documents a change.

Normal contributions do not need to stop. Focused client and provider PRs continue to land as usual. Only changes that introduce or substantially reshape the same boundary currently being migrated may need to coordinate with the relevant step. Where that happens I would rather realign the PR than let it sit.

#496 feeds into step 2. The exact registry or module shape is deliberately not decided here: that is an outcome of defining the contract, not an input to this issue.

[...]

Issue #496, opened 2026-08-23
Proposal: consolidate per-client touch points into a declarative registry


TVick64889:
## Feature Description

Consolidate the 13 hand-edited touch points required to add a new tracked client (per `AGENTS.md`'s own "Adding a tracked client" table) down to a single declarative per-client registry, cutting the manual steps to roughly 4.

## Motivation

- `AGENTS.md`'s "Adding a tracked client" section lists 13 separate files that must all agree on the same client id today, with no mechanical check that catches a missed spot until a test fails or a user reports a client "attributes to nothing."
- `collector.js`'s `clientSourceRoots()` has grown from ~198 lines at v0.1.0 (3 clients) to ~3,785 lines today (32+ clients) across 89 commits, with no shared abstraction ever introduced for how a client plugs in.
- Several recurring WSL2 path-handling bugs (#22, #192, #289, #404, #425, #459, #375) trace back to per-client path knowledge living in at least two independently-maintained places (`clientSourceRoots()` and `WSL_DATA_MARKERS`/`MARKER_CLIENTS`) that must be kept in sync by hand.
- Partial precedent already exists in this codebase: the Limits subsystem (`commandcodeLimits.js`, `copilotLimits.js`, `grokLimits.js`, `zaiLimits.js`, etc.) already uses one file per provider behind `limitCollector.js`'s dispatch. This proposes applying that same pattern to the usage-collection side.

## Proposed Solution

Add a single ordered array (e.g. `src/shared/clients/registry.js`) as the source of truth for per-client metadata (label, vendor, color, icon slug, checkIds, wslMarkers, selfSynced flag). Derive the currently-duplicated lists (`DEFAULT_CLIENTS`, `CLIENT_SOURCE_CHECK_IDS`, renderer maps, Discord RPC maps, CSS icon rules, `SELF_SYNCED_CLIENTS`) from that registry instead of hand-maintaining each independently.

Deliberately **out of scope / left untouched**:
- `clientSourceRoots()`'s actual path-resolution bodies (the hard-won tokscale-mirroring logic) — those stay exactly as fragile/hand-verified as they are today. The registry only adds a guard test cross-checking it against the registry, not a migration.
- `WSL_DATA_MARKERS`/`MARKER_CLIENTS` stays a separate table (genuinely a distinct data source), but gains a guard test so a missing entry fails CI instead of silently "attributing to nothing."
[...]

PR #551, opened 2026-08-29
test: guard WSL marker attribution and Discord Rich Presence client maps

TVick64889:
Follow-up to #496 — scoped to guard-tests-only per your reply there. No registry, no codegen, no file moves.

## What this adds

`tests/shared/clientRegistrationConsistency.test.js`, two invariant checks that weren't covered by the existing `clientTracking.test.js` / `clientHealth.test.js` guards:

1. **`WSL_DATA_MARKERS` \u2194 `MARKER_CLIENTS`** (`src/shared/wslUsage.js`) — every marker has an attribution and vice versa, and every attribution resolves to a real `KNOWN_CLIENTS` id. Currently clean (no drift today), but nothing guarded it going forward — this is exactly the pattern your own doc comment on `WSL_DATA_MARKERS` warns about ("a marker without a `MARKER_CLIENTS` entry attributes to nothing").
2. **`discordRpc.js`'s `KNOWN_CLIENT_ASSETS` / `CLIENT_LABELS` vs. `KNOWN_CLIENTS`** — this pairing had no test at all before this PR.

## Heads up: check #2 is currently RED, left that way on purpose

Running the new test against current `main` turns up a real, live gap: **`qodercn` is a known/tracked client with a `CLIENT_LABELS` entry, but is missing from `KNOWN_CLIENT_ASSETS`.** Per `buildPayload()`'s own logic, that means Discord Rich Presence silently shows no icon (falls through the `KNOWN_CLIENT_ASSETS.has(top)` check) whenever Qoder CN is a user's top client for the day — the label still renders correctly via `CLIENT_LABELS`, just no `smallImageKey`/`smallImageText`.

I deliberately did not fix the one-line omission here, since this PR is scoped to checks over the existing wiring, not wiring changes — happy to add that fix in this PR, a separate one, or leave it to you, whichever you'd rather.

Also noted, not asserted on: `'gemini'` is present in both `KNOWN_CLIENT_ASSETS` and `CLIENT_LABELS` but isn't a `KNOWN_CLIENTS` id — looks like a legacy/ahead-of-launch entry rather than a bug, so I left it alone and didn't add an orphan-detection assertion for it (see code comments — the orphan sets are computed and available for review but intentionally not asserted, since an asset uploaded ahead of a client landing is a legitimate state, not drift).

## What I deliberately did NOT add

- No registry, no codegen (per your answer to open question 1 on #496).
- No renderer `app.js` scraping-based checks — per your note, I didn't want to add another check pinned to `app.js` being a plain, export-less script if it wasn't already load-bearing for something this PR needs. Everything here reads from real `module.exports` (`clientTracking.js`, `wslUsage.js`) or the same sandboxed-`vm` extraction `tests/electron/discordRpc.test.js` already uses for `discordRpc.js`'s `buildPayload` — no new brittle source-pattern assertions.
- No change to `clientTracking.test.js` / `clientHealth.test.js` — their existing coverage (order agreement, source-check-id agreement) isn't duplicated here.

## Testing

- `node --test tests/shared/clientRegistrationConsistency.test.js` — 2 pass, 1 fails red on the qodercn gap above (expected/intentional).
- `npm run lint` — clean.
- Full `npm test` — no new failures beyond the one intentional red test above; the two other failures in the full run (`macWidgetLaunchServicesRecovery.test.js` symlink test, `collectorLoadGuards.test.js` Hermes watchIgnoreMatcher test) reproduce identically on a clean `main` checkout without this PR's file, confirmed by removing it and re-running — pre-existing/environment-specific (Windows dev machine), not caused by this change.

---
## Summary by cubic
Adds invariant consistency checks over the existing per-client wiring to guard against silent drift as those lists evolve, scoped to guard-tests only per #496 — no registry, codegen, or file moves.

- `tests/shared/clientRegistrationConsistency.test.js` verifies that every `WSL_DATA_MARKERS` entry has a `MARKER_CLIENTS` attribution (and vice versa) and that every attribution resolves to a real `KNOWN_CLIENTS` id.
- The same file checks `discordRpc.js`’s `KNOWN_CLIENT_ASSETS` / `CLIENT_LABELS` against `KNOWN_CLIENTS`; this pairing previously had no test at all.
- Each check tests an invariant between two independently-maintained lists, not their current contents, so it keeps protecting the eventual registry migration (#550) as the lists grow.
- The Discord check is intentionally RED against current `main`: `qodercn` has a `CLIENT_LABELS` entry but no `KNOWN_CLIENT_ASSETS` entry, so Rich Presence shows the label but no icon when Qoder CN is the top client for the day. The one-line fix is deliberately left out since this PR is scoped to checks.
- `gemini` appears in the Discord maps but not `KNOWN_CLIENTS`; treated as a legitimate ahead-of-launch entry, so orphan sets are computed but not asserted on.
- Reuses the same sandboxed-`vm` extraction that `tests/electron/discordRpc.test.js` already uses, so no new dependencies or brittle source-pattern assertions.

**Testing**
- `node --test tests/shared/clientRegistrationConsistency.test.js` — 2 pass, 1 intentional red on the `qodercn` gap above.
- `npm run lint` is clean; the two other failures in the full suite reproduce identically on a clean `main` checkout and are pre-existing/environment-specific.

<sup>Written for commit 224cc08b5b6dbd9c492017125d5c410b22236cf8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/551?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


TVick64889:
Heads up on CI: the red `test` jobs are the intentional `qodercn` gap I called out above (`discordRpc KNOWN_CLIENT_ASSETS`/`CLIENT_LABELS` check) — not a regression, that's the guard test proving it catches the real drift. Everything else is green.

Keeping this PR scoped to checks only per your note. Whenever you're ready, the one-line follow-up (`'qodercn'` → `KNOWN_CLIENT_ASSETS` in `src/electron/discordRpc.js`) is ready to go as its own small PR — just say the word and I'll open it.

Javis603:
Thanks @TVick64889. Both checks are the shape I was after, and reusing the sandboxed `vm` extraction from the existing Discord test instead of adding another source-scraping assertion is exactly the kind of solution I had in mind. Keeping the wiring fix out of a test-only PR was the right call too.

I pushed the `qodercn` fix to main separately (b89dce73) rather than taking it here, since the omission is mine and this PR should stay within the scope I asked for. CI against the updated base is green, and the two failures you isolated do not reproduce there either, including on `windows-latest`, so they look specific to your local setup rather than platform-related.

On `gemini`: it was never a tracked client. It has not appeared in `DEFAULT_CLIENTS` or `KNOWN_CLIENTS` at any point in the repo's history, so the entries in `app.js` and `discordRpc.js` are unwired leftovers from the initial release.

Leaving the reverse orphan direction unasserted was still the right choice for this PR. The safety contract we need here is that every tracked client has the required Discord metadata; deciding whether extra presentation entries should be rejected or cleaned up is a separate constraint and is not required for this guard.

PR #566, opened 2026-08-31
fix: seed initial limit providers from detected tools

### LINKED ISSUE #537: Seed AI Limits providers from detected tools on first run
## Problem

AI Limits currently enables all providers by default, which makes the initial experience noisy for users who only use a few AI tools.

I don't think we should solve this by dynamically hiding providers based on runtime states such as `notConfigured` or `noSyncedData`. Once a provider has been enabled, its visibility should remain user-controlled even if credentials expire, discovery temporarily fails, or synced data is unavailable.

## Proposed behavior

On a true fresh install only, use the existing Collection tool discovery results to seed the initial AI Limits provider selection.

After Collection discovery has completed:

- If an AI tool has an actually detected local source/path and maps to an AI Limits provider, enable that provider by default.
- If no corresponding tool/source was detected, leave that Limits provider disabled.
- Collection tools without a corresponding Limits provider should be ignored.
- Limits providers without a corresponding Collection tool should remain disabled by default.

For example:

```text
Detected by Collection:
Claude Code     ✓
Codex           ✓
OpenCode        ✓
Cursor          ✓
Antigravity     ✓
Hermes Agent    ✗
Cline           ✗
Qwen            ✗

Initial AI Limits selection:
Claude          ✓
Codex           ✓
OpenCode        ✓
Antigravity     ✓
Other providers ☐
````

The Collection setting being enabled is not enough to count as detection, since Collection tools are enabled by default. There should be an actual detected source/path.

## Important behavior

This initialization must happen **only once, on a genuine first run**.

After the initial selection has been persisted, AI Limits provider selection must be entirely user-controlled.

The following must **never** automatically change the selection afterward:

* restarting the app
* updating Token Monitor
* installing or removing an AI tool
* later Collection discovery results
* a previously detected path becoming unavailable
* credentials expiring or being removed
* API/probe failures
* missing synced data
* adding a new Limits provider in a future Token Monitor version

In other words:

> Discovery informs the initial defaults; it does not control ongoing provider state.

Existing installations upgrading to a version with this behavior must keep their current provider selection unchanged.

## Implementation notes

The first-run decision should be captured from persisted app state at startup, before defaults or migrations can affect it. It should not be inferred from the absence of a newly added settings field, since existing installations would also lack that field after an upgrade.

Initialization should wait until Collection discovery has completed before persisting the initial provider selection.

If no supported tools are detected on the first completed discovery, seed Codex as a narrow fallback so the AI Limits view remains discoverable. This fallback applies only to automatic first-run initialization; an explicitly empty user or environment selection must remain empty rather than falling back to all providers.

Related to #54. This is intended to address the initial provider clutter without continuously hiding enabled providers based on runtime state.

mikemikimike:
## Summary

- Seed AI Limits providers once on a genuine first run from Collection clients whose local source is detected.
- Fall back to Codex only when the first completed discovery finds no supported local source, keeping AI Tool Limits discoverable.
- Preserve existing installations, explicit `TOKEN_MONITOR_LIMIT_PROVIDERS` configuration, and subsequent user changes.
- Apply the persisted seed immediately to the running Limits runtime and renderer.
- Treat an explicitly empty provider configuration as no providers, while an omitted configuration continues to enable all supported providers.

## Verification

- `npm run verify` passed locally.
- Focused first-run, source-health, runtime-reconfiguration, configuration, and provider-order regression tests passed.
- Fresh-profile visual verification confirmed that the updated implementation selects only detected tools instead of enabling every provider.

Closes #537

---
## Summary by cubic
Seeds AI Limits providers on a genuine first run from the first completed discovery snapshot, using only providers whose local client source is currently `detected` in source health, and documents `TOKEN_MONITOR_LIMIT_PROVIDERS` semantics (omitted = all providers, explicitly empty = none).

- Runs once per installation across all collector modes, and retries if the settings write fails.
- Skips seeding when the settings file already exists or `TOKEN_MONITOR_LIMIT_PROVIDERS` is set, preserving existing installs, explicit overrides, and pre-discovery user changes.
- Falls back to Codex when no local source is detected so the Limits view stays discoverable.
- Applies the persisted seed to the running Limits runtime and renderer.
- Maps detected Collection client aliases (`micode` → `mimo`, `zcode` → `zai`, `qodercn` → `qoder`) to their Limits provider ids.
- Syncs the shared provider helper to the worker copy and refreshes the hub build registry to revision 24 in both copies.
- Updates `.env.example` and `docs/configuration.md` to match the new env semantics and restores the provider order reference.

| Area | Before | After |
| --- | --- | --- |
| New-install limit providers | Not seeded from detected tools | Seeded from sources whose local client is `detected` |
| Source-free first run | Limits unavailable | Falls back to Codex |
| Explicitly empty `TOKEN_MONITOR_LIMIT_PROVIDERS` | Treated as "all providers" | Treated as "no providers" |
| Existing installs / explicit `TOKEN_MONITOR_LIMIT_PROVIDERS` | Unchanged | Unchanged; seeding skipped |

**Verification:** Lint and focused Node tests passed covering the new helper (including client alias mapping), first-run wiring, source health, and runtime config. The full suite stalled downloading the Electron binary, with no failures reported before the stall.

<sup>Written for commit 5df1202a1c0f395a38d18561573ff1821fdf172f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/566?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Javis603:
Thanks @mikemikimike for taking the first pass at #537 and putting the one-time first-run seeding structure in place.

I took over the branch afterward and made several maintainer follow-ups before merging. The original implementation still had a few end-to-end correctness gaps — in particular, an empty detected set could round-trip back to “all providers enabled”, and the initial seed was not yet using the authoritative source-presence signal or updating the running Limits runtime.

The follow-up changes now:

- seed only from actually detected Collection sources, including the client/provider aliases;
- preserve a real empty provider selection instead of falling back to all providers;
- apply the persisted seed immediately to the running Limits runtime and renderer;
- roll back and retry cleanly if persistence fails;
- preserve existing installs and later user choices;
- fall back to Codex only for a source-free fresh install so AI Tool Limits remains discoverable.

The related configuration docs, worker copy, registry, and regression coverage have also been updated.

`npm run verify` passes locally, and CI + CodeQL are all green. The final behavior now matches #537: discovery only seeds the initial defaults once, and provider selection is entirely user-controlled afterward.

Thanks again for contributing the initial implementation. I'm going to squash and merge this.

PR #569, opened 2026-08-31
feat(qodercn): cloud credits usage adapter for the CN desktop client

shuming9315-Fuxiao:
## Summary

The Qoder CN desktop client (`com.qodercn.app.stable`, 0.1.x rewrite) stopped persisting token-level usage locally: the legacy `%APPDATA%\QoderCN\...\local.db` has been stale since 2026-08-26, and the new `main.sqlite` stores chat content only — `chat_session_context_usage.totalTokens` is always 0 and `chat_session_messages` payloads carry no usage object. As a result the existing `qodercn` client rows froze at the old install.

The account quota API (`GET /api/v2/quota/usage` on `openapi.qoder.com.cn`) is the only live source: it answers cumulative credits (main quota + package quotas + plan tier). This PR adds a cloud adapter that reads it and derives daily usage by snapshot differencing.

## Changes

- **`src/shared/qoderCnCloudUsage.js`** (new)
  - Automatic `dt-` access token: DPAPI decrypts the AES key from the client's Local State, AES-256-GCM then decrypts `%APPDATA%\com.qodercn.app.stable\auth.v1.dat`. An explicit token via settings/env wins when present.
  - `GET /api/v2/quota/usage` with `Cosy-ClientType: 10` + `User-Agent: Qoder` headers (required by the endpoint), 12s timeout, 5-minute response cache so watch-triggered ticks from other clients do not hammer it.
  - Daily usage by snapshot differencing: state file `qoder-cn-credit-snapshots.json` (per-account, 400-day retention); a decreasing cumulative value (reset/renewal/new package) rebaselines instead of emitting negative usage.
  - `fetchQoderCnCloudLimits()`: projects the quota into the limits-provider shape (main credits window + one window per package).
- **`src/shared/qoderLimits.js`**: when `qoderSite=cn` and no web cookie is configured, falls back to the cloud limits. `MISSING_TOKEN` returns the quiet `notConfigured` state; network errors return `unavailable` with a visible diagnostic. Global-site accounts behave exactly as before.
- **`src/shared/collector.js`**: the periodic collection point and the history-chart integration point merge the cloud credit rows into the qodercn rows. Cloud failures are non-fatal — the legacy local-db history still reports.
- **`src/shared/qoderCnUsage.js`**: map the `qmodel_38max` model code added in the 0.1.2 client to `Qwen3.8-Max-Preview` (same preview model as the retired `qmodel_preview`).

## Verification

- Token auto-decryption: OK (no settings input required)
- Quota read: main credits 1557/2000, package `Qwen 专属积分` 500/500 (exhausted, `available=false`), plan tier `Pro`
- Snapshot differencing: first run establishes the baseline (0 rows); subsequent ticks emit only the delta; same-quota re-read emits 0 rows
- Limits projection renders as `Qoder | Pro | Credits 23% left | 443/2000 | resets in 25d` plus the package window
- `qoderSite=global` without a cookie: `notConfigured` (upstream default behavior preserved)
- Module load + syntax check against upstream `main`: clean (only the four files above differ)

## Notes

- Cloud token-level usage is not available: `/api/v1/me/usages/big_model_credits/histories` → 404, `/api/v2/...` → 503 at the ALB layer, and a scan of the client's bundled JS found no other history endpoint. Daily usage is therefore expressed in credits.
- Closes the gap left by #509 (which added a manual dt- token input for the CN site): this adapter reads the token from disk automatically, so no manual input is needed. Happy to rebase/combine if preferred.

---
## Summary by cubic
Adds a cloud usage adapter for the Qoder CN desktop client so `qodercn` rows stay live after the 0.1.x client rewrite stopped persisting token-level usage locally (`local.db` stale, `main.sqlite` always 0).

- New `src/shared/qoderCnCloudUsage.js` auto-decrypts the client's `dt-` token from disk (DPAPI + AES-256-GCM on Windows), queries `GET /api/v2/quota/usage`, and derives daily credit usage via snapshot differencing (baseline on first run, 400-day retention, rebaselines on decreasing totals instead of emitting negative usage).
- `qoderLimits.js`: the CN site without a web cookie now falls back to cloud limits; a missing token stays in the quiet `notConfigured` state and network errors surface as `unavailable`.
- `collector.js` merges cloud credit rows into `qodercn` rows at both collection points; cloud failures are non-fatal and legacy history still reports.
- `qoderCnUsage.js` maps the 0.1.2 client's `qmodel_38max` code to `Qwen3.8-Max-Preview`.

| Area | Before | After |
| --- | --- | --- |
| `qodercn` usage data | Frozen at the old install (stale local DB, `main.sqlite` carries no usage) | Live daily credit usage from the cloud quota API via snapshot differencing |
| CN site auth | Manual `dt-` token input required when no web cookie exists | Token auto-decrypted from the client's `auth.v1.dat`; an explicit token/env still wins |
| CN limits view | `notConfigured` | Credits and per-package windows with plan tier label (e.g. `Pro`, credits meter) |
| Model `qmodel_38max` | Unmapped in 0.1.2 client data | Mapped to `Qwen3.8-Max-Preview` |

Implementation notes: token auto-decryption is Windows-only — other platforms need an explicit token; the DPAPI call uses `koffi` dynamically. A rejected token (401/403) clears the auth cache. The API exposes no token-level usage (only cumulative credits), so daily usage is expressed in credits and usage while the monitor was off is not recoverable.

Verified: token decrypt, quota read (main 1557/2000, exhausted package, `Pro` tier), differencing (first-run baseline, delta-only ticks, zero on same quota), limits projection, and global-site behavior unchanged.

<sup>Written for commit 4a04c95829f75b5b0ab9343aee7b1792554a7d73. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/569?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/shared/collector.js
+++ b/src/shared/collector.js
@@ -50,6 +50,7 @@ const {
   qoderCnDataPaths,
   resolveQoderCnPricing
 } = require('./qoderCnUsage');
+const { collectQoderCnCloudRows } = require('./qoderCnCloudUsage');
 const { resolveReasonixStatsDir, REASONIX_SOURCE_CHECK_ID } = require('./reasonixPaths');
 const { resolveDshSessionsDir, DSH_SOURCE_CHECK_ID } = require('./dshPaths');
 const { indexDshSessionHeaders, readDshSessionHeader, resolveDshSessionsRoot } = require('./dshSessionFiles');
@@ -1572,6 +1573,21 @@ async function collectUsageOnce(options) {
       try {
         const qoderCnSinceMs = anchorUsed ? new Date(collectedAt.getFullYear(), collectedAt.getMonth(), collectedAt.getDate()).getTime() : undefined;
         qoderCnRows = await collectQoderCnRows({ homeDir: options.homeDir, logger: options.logger, sinceMs: qoderCnSinceMs });
+        // The 0.1.x Qoder CN client keeps no local token-usage database, so the
+        // legacy rows above stop at the old install. Cloud credits keep the
+        // qodercn row live: quota snapshot-diff rows, merged in the same shape.
+        // Cloud failures are non-fatal — the legacy history still reports.
+        try {
+          const qoderCnCloudRows = await collectQoderCnCloudRows({
+            homeDir: options.homeDir,
+            logger: options.logger,
+            sinceMs: qoderCnSinceMs,
+            nowMs: collectedAt.getTime()
+          });
+          if (qoderCnCloudRows.length) qoderCnRows = qoderCnRows.concat(qoderCnCloudRows);
+        } catch (cloudErr) {
+          if (typeof options.logger === 'function') options.logger(`qodercn cloud credits skipped: ${cloudErr.message}`);
+        }
         qoderCnPricing = await resolveQoderCnPricing(qoderCnRows, {
           lookupModelPricing: options.lookupModelPricing || lookupModelPricing,
           commandTimeoutMs: options.pricingTimeoutMs,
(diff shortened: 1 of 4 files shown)


shuming9315-Fuxiao:
Follow-up after running both locally: keeping the daily usage rows credit-denominated breaks the token-unit consistency of the history panel (every other client reports tokens, and mixing credits into the same daily/monthly totals was confusing on my dashboard). #576 replaces the usage rows with transcript-based token estimates and keeps this PR quota/limits card as the authoritative credits view.

If both are desirable, the usage-row merge hunks in this PR could be dropped so the two paths do not double-count — happy to rebase either way.

Javis603:
Thanks for validating the CN quota endpoint, authentication path, and snapshot behavior. After comparing the approaches, I am keeping credits in AI Tool Limits and token/session history in token units. This PR mixes credit deltas into `qodercn` usage rows, so I am not carrying that collector path forward.

The reusable limits and authentication work overlaps #509, while current local token data is being handled in #753. I am closing this one so those two responsibilities stay separate.

PR #581, opened 2026-09-01
feat(pricing): price first-party models the catalogs do not carry yet

krudo-taco:
## Summary

Claude Fable 5.1 (`claude-fable-5-1`, released this week) shows tokens but $0 everywhere in Token Monitor. tokscale prices rows from the LiteLLM / OpenRouter / models.dev catalogs it downloads (1h TTL) and keeps a row it cannot price at `cost: 0`; the bare `claude-fable-5-1` key is in none of the three (OpenRouter spells it `anthropic/claude-fable-5.1`, models.dev `anthropic/claude-fable-5-1`, LiteLLM has nothing yet), so until a catalog picks the model up every Claude Code row for it is unpriced, and the same gap reopens on every first-party release.

This adds `src/shared/modelPricingFallback.js`: a small bundled table of official list rates (USD per million tokens; Fable 5.1 at $10 / $50, cache read $0.25, cache write $12.50, cross-checked against the OpenRouter and models.dev rows) and pure functions that re-cost a tokscale row only when all of these hold: tokscale returned `cost: 0`, the row's `provider` and canonical model id match a table entry exactly, the row has billable tokens, and the model is not a key of tokscale's `custom-pricing.json` (matched the way tokscale matches it, lowercased and exact, so a `$0` the user wrote there stays a deliberate "free"). Rows tokscale priced, other providers, other models and zero-token rows come back by reference, untouched.

`collector.js` wraps the scan runner in `collectUsageOnce` and the graph runner in `collectHistoryOnce`, so full scans, the anchored `--today` watch tick (the delta stays exact because both sides are costed the same way), the WSL scan (which borrows `runTokscaleFn`), history and injected test runners all inherit it. The custom-priced id set is read once per tick and re-read when `custom-pricing.json` changes; one log line per model per process says the bundled rate was used. The table's revision is folded into `configFingerprint`, so the first start after this change (and after any future table edit) does one full scan instead of deriving month/allTime from a persisted anchor that was costed under the old table — otherwise the stale $0 would survive until the hourly full scan. No wire-shape change, no new dependency, `usage.js` untouched, so no hub build registry bump.

## Verification

- `npm run verify` (lint + the full `node --test` suite): 3962 tests, 3961 passed, 0 failed, 1 skipped (the pre-existing skip), lint clean. One earlier run on the same tree hit a timing flake in `tests/shared/watcherHost.test.js` ("no event from the worker" after 15 s in the real-worker reconfigure test); it passes in isolation and in the other full runs, and nothing here touches the watcher.
- New tests: `tests/shared/modelPricingFallback.test.js` (table shape and a revision guard, canonical ids, provider/model gating, custom-key semantics, immutability, scan and graph shapes) and `tests/shared/collectorPricingFallback.test.js` (full scan, anchored watch tick, WSL runner, history graph, custom-pricing exclusion and re-read, log dedupe); `tests/shared/collectorAnchorPersistence.test.js` updated for the fingerprint segment.
- `tokscale graph` rows carry `providerId` in 4.15.0 (checked on real output, alongside `provider` on scan rows), which is what the history path keys on.
- End-to-end on macOS with tokscale 4.15.0: copied the real catalog cache into a scratch `TOKSCALE_CONFIG_DIR`, deleted every `fable-5.1` / `fable-5-1` key, and pointed `HTTPS_PROXY` at a dead port so tokscale degraded to that cache. Raw `tokscale --json --client claude --today` then reports `claude-fable-5-1 anthropic cost=0`, while `collectUsageOnce` and `collectHistoryOnce` through the real spawn path report `today.modelCosts["claude-fable-5-1"] = 13.378543` and the identical per-model cost on the history day, with the fallback logged once. With the untouched current cache tokscale prices the model itself and the fallback stays inert; a unit test pins the two to the same number for one real row (2.6361165).

## Notes

- The table deliberately holds only Fable 5.1, the model the catalogs lack today. `claude-mythos-5-1` shares the $10 / $50 rate but its cache-read rate is not published, so it is left out rather than guessed. Adding a row is one line plus the checklist in the module header (and a revision bump the guard test enforces); a row can be dropped once LiteLLM carries the bare key.
- tokscale infers `provider` from the model name, so the table is keyed on first-party ids billed at list price only. That is why the module header says not to add reseller or subscription ids, and to keep one rate per model regardless of context length.
- `cacheWritePerM` is the 5-minute cache-write rate, the only one the catalogs carry and therefore the one tokscale applies; 1-hour writes are under-costed by both alike.
- No existing issue covers this. #243 (enforcing custom overrides inside Token Monitor) touches the same area but is orthogonal: this change never overrides a price, it only fills a `$0`.

---
## Summary by cubic
Prices first-party models the pricing catalogs have not picked up yet (e.g. this week's `claude-fable-5-1`) from bundled list rates, so their tokens no longer show as $0 in Token Monitor.

- Adds `src/shared/modelPricingFallback.js` with list rates for `claude-fable-5-1` ($10 / $50 per M tokens, cache read $0.25, cache write $12.50).
- Re-costs scan and history-graph rows only when tokscale reported `cost: 0`, the provider and canonical model id match the table, and the row has billable tokens.
- Wraps the scan and graph runners in `collector.js`, so full scans, anchored watch ticks, WSL scans, and history inherit the fallback.
- Skips models present in `custom-pricing.json` (matched the way tokscale matches them), so a user-set price — including an explicit `0` — always wins.
- Folds the table revision into `configFingerprint`, so an edit or first start after this change forces one full scan instead of reusing an anchor that carries stale $0 rows.

| Area | Before | After |
| --- | --- | --- |
| `claude-fable-5-1` (and future first-party releases) | Tokens show but cost is $0 everywhere until a catalog carries the bare key | Costed from bundled list rates; falls back to tokscale once a catalog covers the model |
| First start after upgrade or table edit | Warm ticks reuse the persisted anchor, so stale $0 rows survive until the hourly full scan | One full scan retires the anchor; warm ticks continue as before |

**Implementation notes**
- The table deliberately holds only `claude-fable-5-1`; `claude-mythos-5-1` shares its input/output rates, but its cache-read rate is unpublished, so it is left out rather than guessed.
- Cache write applies tokscale's own 5-minute rate — the only one the catalogs carry.
- The fallback is keyed on first-party ids billed at list price; reseller and subscription ids are out of scope by design.

**Tests and validation**
- `npm run verify` passes (3961/3962 tests; the single failure is a pre-existing `watcherHost` flake that passes in isolation).
- New tests cover table shape and revision guard, provider/model gating, custom-key semantics, immutability, and every collector path (scan, watch tick, WSL, history, custom-pricing re-read, log dedupe).
- End-to-end on macOS with a stale catalog cache: `tokscale` reports $0 while both collectors report the bundled rate; with a current cache the fallback stays inert.

No wire-shape change, no new dependency, and no hub build registry bump required.

<sup>Written for commit 0725ef86d772df86f9e4d891c21f9668df5f8cd2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/581?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Javis603:
Thanks @krudo-taco for the detailed write-up and for testing this end to end against a real catalog cache.

I checked the premise against tokscale 4.15.0 before deciding, and I can't reproduce the $0. `tokscale pricing claude-fable-5-1` resolves to an exact LiteLLM key at the same rates your table carries ($10 / $50, cache read $0.25, cache write $12.50). Removing every fable key from the LiteLLM cache doesn't produce a $0 either: tokscale then matches OpenRouter's `anthropic/claude-fable-5.1` through its model-part resolution, at the same rates.

More generally, I don't want to maintain bundled pricing in Token Monitor. Custom model pricing already exists through tokscale's `custom-pricing.json` (101a1e2), with a settings UI under Settings → Collection → Custom model pricing (a01d2e7). That already covers a model that is temporarily missing or mispriced upstream, and it applies right away instead of waiting on a code change, a release, and an app update.

image

A bundled table would be a second pricing source needing an edit for every future model launch or price change. In this case, the catalogs already carry the model, while this fallback has not yet reached a release.

I'm closing this one, but thanks for the work that went into it.

PR #585, opened 2026-09-02
feat(models): rank models by tokens or cost

JulianZJN:
## Summary

- add a persisted **Model ranking** setting with `Token usage` and `Cost` options
- keep token ranking as the compatibility default while letting cost ranking drive both model order and bar length
- keep the Home model summary token-ranked and leave every non-model breakdown unchanged

## Behavior

When `Cost` is selected, models with known costs are sorted by cost descending. Equal costs use token volume as the tie-breaker, and models without cost attribution follow priced models. If a period has no usable model-cost data, the list and bars fall back to token volume instead of becoming empty.

Token and cost values remain visible in both modes. The setting is normalized in the main process, persisted with the existing settings flow, and translated in every bundled locale.

## Validation

- `npm run verify` — 3965 passed, 1 skipped, 0 failed
- packaged and launched the arm64 macOS app locally with ad-hoc signing
- verified the setting persists and the TOTAL model list is ordered by descending cost

### Setting

Model ranking setting

### Cost-ranked model list

Models ranked by cost

---
## Summary by cubic
Adds a configurable Model ranking setting so the Models list can be ordered and scaled by cost instead of token usage.

- Adds a persisted `Model ranking` setting with `Tokens` and `Cost` options; token ranking remains the default.
- When `Cost` is selected, models with known costs sort by cost descending; models without known cost data follow priced ones, and token volume breaks ties.
- Falls back to token volume for order and bars when no model has known cost data in the period.
- Leaves the Home model summary token-ranked and all non-model breakdowns unchanged.
- Normalizes the setting in the main process, persists it through the existing settings flow, and translates it in every bundled locale.
- Moves bar scaling and width into `breakdownRenderPolicy` so cost-ranked bars scale to the largest rendered value; fractional USD amounts no longer collapse against a minimum scale of 1.
- Uses the same `Tokens` / `Cost` inline control as the heatmap metric toggle and places it below the currency rate row.

| Area | Before | After |
| --- | --- | --- |
| Models ranking | Models are always sorted and scaled by token usage. | Models can be sorted and scaled by cost when `Cost` is selected; falls back to tokens when cost data is missing. |
| Settings | No model ranking option. | New `Model ranking` setting with `Tokens` and `Cost` options. |

**Validation**

- `npm run verify` passes: 3965 passed, 1 skipped, 0 failed.
- Packaged and launched the arm64 macOS app locally with ad-hoc signing; verified the setting persists and the Models list orders by descending cost.

<sup>Written for commit 67dce924e4baa5247b454d84fb318535ceec7361. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/585?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Javis603:
Thanks @JulianZJN for this! I pushed a few small fixes to your branch: cost-ranked bars were scaling against a hardcoded minimum of 1, which made costs below $1 render at the wrong proportions. I also aligned the new setting and its wording with the heatmap’s existing Tokens/Cost toggle, and cleaned up a couple of related settings/UI details.

PR #609, opened 2026-09-04
fix(renderer): leave settings overlay when switching views via the switcher

### LINKED ISSUE #607: View switcher does not switch views while the settings panel is open
### What happens

Open the settings panel (gear button), then switch views with the view switcher in the bottom-left footer. The switcher updates its label, but the UI stays on the settings panel — the view behind it never changes.

Switching views from the tray menu works correctly.

### Why

`renderBreakdownChange()` calls `render()`, but while the settings panel is open `visibleStatsSurface()` reports `settings`, so `render()` returns early before painting the new view. The tray path (`openViewFromTray`) hides the settings panel first, but the view-switcher click handlers (both the current-view button and the menu items) call `renderBreakdownChange()` without doing that cleanup, so the panel stays open and blocks the repaint.

### Steps to reproduce

1. Click the gear button to open settings.
2. Click the view switcher (bottom-left) and pick a different view, or click the switcher to cycle to the next view.
3. The switcher label changes, but the settings panel remains on screen and the view does not switch.

### Environment

- Token Monitor 0.53.0
- Windows 11 ARM64

qwertlooker:
Fixes #607.

## Summary

Switching views with the bottom-left view switcher did nothing while the settings panel was open: the switcher label updated, but the UI stayed on the settings panel.

While the settings panel is open, `visibleStatsSurface()` reports `settings`, so `render()` returns early before painting the new view. The tray path (`openViewFromTray`) already leaves the settings overlay before calling `renderBreakdownChange()`, but the view-switcher click handlers (both the current-view button and the menu items) call `renderBreakdownChange()` directly, so the panel stayed open and blocked the repaint.

This change moves that cleanup into `renderBreakdownChange()` itself: when the settings panel is open, it is hidden (and the list search / shortcut-recording state reset, same as `openViewFromTray`) before rendering. This also keeps the two entry points from drifting apart, mirroring the invariant elsewhere in the codebase that a single function owns each behavioural rule.

## Behaviour change

- View switcher (button click and menu items) now exits the settings overlay and switches views, matching the tray behaviour.
- No changes to settings persistence or to the tray path's own handling.

## Verification

- `npm run lint` — clean.
- `node --test "tests/electron/*.test.js"` — 1668 pass, 1 fail. The single failure (`macWidgetLaunchServicesRecovery`) is pre-existing and environment-specific: creating a symlink without admin/developer-mode privileges on Windows fails with `EPERM` before any test logic runs; it does not touch this change.
- Full `npm test` was also started locally but exceeds a reasonable local run time on this Windows ARM64 machine; CI covers the full matrix on Node 22 & 24.

---
## Summary by cubic
Fixes #607: switching views via the view switcher while the settings panel is open now repaints the main view behind the settings overlay instead of leaving the UI stuck on the old content.

- Settings is no longer a separate stats surface; `visibleStatsSurface()` only reports `null`, `'bubble'`, or `'main'`, and settings visibility is tracked separately.
- Stats updates always repaint the main view and refresh the settings panels when the overlay is open; the `statsRenderScheduler.clear()` path is gone so pending repaints flush even with settings visible.

| Area | Before | After |
| --- | --- | --- |
| View switcher with settings open | Label updated, main view never repainted | Main view repaints behind settings; settings stays open |
| Stats while settings open | Main view not repainted until settings closed | Main view stays live behind the overlay |

- The tray path and settings persistence are unchanged.
- Lint passes; the only test failure is the pre-existing Windows-specific `macWidgetLaunchServicesRecovery` case, unrelated to this change.

<sup>Written for commit 77a63bb614f50360d16be1904127408d73d02f31. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/609?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/electron/renderer/app.js
+++ b/src/electron/renderer/app.js
@@ -6386,13 +6386,14 @@ function isRendererWindowHidden() {
 function visibleStatsSurface() {
   return statsRenderSchedulerApi.visibleStatsSurface(
     isRendererWindowHidden(),
-    state.floatingBubble.collapsed,
-    isSettingsPanelOpen()
+    state.floatingBubble.collapsed
   );
 }
 
 function isSettingsSurfaceVisible() {
-  return visibleStatsSurface() === 'settings';
+  return !isRendererWindowHidden()
+    && !state.floatingBubble.collapsed
+    && isSettingsPanelOpen();
 }
 
 function serviceStatusSurfaceVisible() {
@@ -8061,15 +8062,6 @@ function setBreakdown(breakdown, options = {}) {
 
 function renderBreakdownChange(breakdown, options = {}) {
   if (!setBreakdown(breakdown, options)) return false;
-  // Switching views always leaves the settings overlay. While the panel is
-  // open, visibleStatsSurface() reports 'settings' and render() returns early,
-  // so the view switcher would update its label without repainting the view.
-  if (isSettingsPanelOpen()) {
-    resetSettingsListSearch();
-    stopWindowShortcutRecording();
-    els.settingsPanel.classList.add('hidden');
-    els.shell.classList.remove('settings-open');
-  }
   state.animateBarsFromZero = true;
   state.animateChartsOnRender = true;
   let renderSucceeded = false;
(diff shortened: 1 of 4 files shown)


Javis603:
Thanks @qwertlooker for tracing #607 to the Settings render gate and putting together the initial fix.

I pushed a follow-up commit (77a63bb6) that keeps Settings open by treating it as an overlay. View changes and live stats now continue rendering behind it, while hidden-window coalescing, floating-bubble isolation, and tray navigation retain their existing behavior.

PR #612, opened 2026-09-04
fix(claude): guard CLI fallback and quota parsing

RSicol:
## Summary

- require `claude auth status --json` to positively confirm an existing login before running the `/usage` quota fallback, preventing background refresh from starting an unauthenticated Claude session
- invoke Claude `.cmd` through explicit `cmd.exe` with shell mode disabled on Windows
- fix CLI quota parsing so missing session or weekly values cannot borrow percentages or reset metadata from another window, including model-specific weekly sections and compact or colon-delimited reset lines

## Scope

- leave existing OAuth credential discovery and refresh behavior unchanged, including native and WSL `.credentials.json` files, macOS Keychain, Windows Credential Manager, and the macOS delegated `/status` refresh path

## Validation

- `npm run verify`: 4,124 passed, 1 skipped, 0 failed
- `git diff --check`
- manually verified that Keychain-backed Claude Pro quota remains detected on macOS

Refs #265


Javis603:
Thanks @RSicol for identifying the unattended Claude authentication path and starting this fix. The final implementation preserves the authenticated CLI `/usage` fallback, gates it with the non-interactive `claude auth status --json` check, and leaves the existing OAuth credential discovery and refresh behavior unchanged.

I pushed follow-up commits for complete PTY capture, duplicate prompt-input handling, section-safe quota parsing, and Windows `.cmd` execution. I also verified the fallback end to end with Claude Code 2.1.263 on macOS, including both session and weekly windows.

PR #621, opened 2026-09-05
feat(watcher): use native fs.watch recursive on macOS to reduce memory

Cheddaran:
# feat(watcher): use native `fs.watch({recursive: true})` on macOS to reduce memory

## Motivation

chokidar's per-directory watch model pinned the installed app's native allocator high-water at ~2.2 GB of compressed/swapped memory on macOS for any user with a multi-thousand-dir tree (OpenClaw's `~/.openclaw/agents/` is ~15k dirs / ~116k files; Claude Code with many projects hits it too). The leading cause is one watch descriptor per directory inside the tree; on a 10k-dir tree chokidar's close() takes ~12 s, and the per-watch native allocation never gets released because new roots are reattached before the old teardown settles.

Switching to `fs.watch(root, { recursive: true })` on darwin uses one FSEvents stream per top-level root instead of one handle per directory inside it. The downstream consumer (collector reaction logic) sees the same `('all', event, filePath)` shape and is unchanged.

**Measured against a real OpenClaw tree on a real macOS install:**

| Metric | Before | After |
|---|---|---|
| Compressed/swapped total | ~2.2 GB | — |
| Total RSS across all processes | — | **625 MB** |
| Main process RSS | — | 131 MB |
| Main process footprint | — | 126 MB |

## Mechanism

- **`src/shared/nativeWatcher.js`** (new). `createPlatformWatcher` picks chokidar vs native based on `shouldUseNativeWatcher`. Both backends surface a chokidar-shaped `('all', event, filePath)` event API; the worker's wire-up is unchanged.
- The watch ignore matcher (`watchIgnoreMatcher`) is applied **post-event** for the native path (fs.watch has no `ignored` option) and at-watch-creation for chokidar. The matcher's pure-function contract keeps the post-event application semantically equivalent.
- FSEvents does not distinguish add/change/unlink in `eventType` (`rename` covers both, `change` covers content); a synthetic `'change'` is emitted. The event type is only used in a log-string downstream, so this is invisible to the consumer; the comment in `nativeWatcher.js` calls it out so a future caller that needs the distinction knows to stat first.
- **`src/shared/watcherWorker.js`** uses `createPlatformWatcher`. `closeCurrent` shape is unchanged; the native backend's close resolves synchronously, so the await is cheap.
- **`src/shared/watcherHost.js`** in-process host stays on chokidar deliberately:
  - The collector's watch-behaviour tests stub `chokidar.watch` to control event delivery. Routing the in-process host through `createPlatformWatcher` would bypass that stub.
  - The in-process host is also the production worker-fallback path when the worker module load fails. Sticking with chokidar there means a worker-failure path uses the well-trodden backend rather than swapping mid-degraded-run.
  - The macOS optimisation is in the worker only.

## Behaviour matrix

| Platform | `usePolling` | Backend |
|---|---|---|
| macOS | `false` (default) | `fs.watch({recursive: true})` (FSEvents) |
| macOS | `true` (env override or sticky ENOSPC fallback) | chokidar |
| Linux | any | chokidar |
| Windows | any | chokidar |

`fs.watch({recursive: true})` on Linux is documented as not fully supported, and chokidar is the only backend that gives us atomic-write stability (`awaitWriteFinish`) on Windows — so the platform split is deliberate, not a future-cleanup item.

`TOKEN_MONITOR_WATCH_POLLING` still overrides in both directions, including forcing the macOS native path back to chokidar when set.

## Tests

- **All 4081 existing tests pass unchanged** (`npm run verify`, lint clean). The chokidar test surface is intact because the in-process host and the test helper (`tests/helpers/watchHost.js`) still go through chokidar.
- **New `tests/shared/watcherNativeRecursive.test.js`** (darwin-only, real fs):
  - `shouldUseNativeWatcher` selects the native path on macOS with no polling, chokidar when polling is requested.
  - delivers a `change` event for a file in a deeply nested subdirectory (`a/b/c/session.jsonl`), proving FSEvents recursive delivery works through `createPlatformWatcher`.
  - applies the ignore matcher post-event — an event from an out-of-clients root is filtered, an event from a tracked root is forwarded.
  - `close()` resolves cleanly so the worker's `closeCurrent` await doesn't hang.

CI on Linux/Windows will skip these tests (the `process.platform === 'darwin'` guard is at test level, not file level), which is the right shape: chokidar end-to-end is already covered by `watcherNativeEvents.test.js` on those platforms.

## Manual verification (on this branch, on macOS)

1. `bash ~/token-monitor-local/sync-local.sh` (or the equivalent manual steps: bump version, apply `claude-model-key-fix.patch`, `npm run update:hub-build`, build in `/tmp`, sign, install to `/Applications`).
2. Launch `/Applications/Token Monitor.app` against a tree with many watched roots (e.g. OpenClaw's `~/.openclaw/agents/`).
3. RSS settles around 600 MB total across all processes (was 2.2 GB). Main process footprint around 126 MB.
4. Drop a file into a watched root; the widget updates within seconds — proves event delivery through the native path.

## Files

- `src/shared/nativeWatcher.js` (new, 134 lines)
- `src/shared/watcherWorker.js` (modified)
- `src/shared/watcherHost.js` (modified — comment-only, explains why in-process stays on chokidar)
- `AGENTS.md` (modified — note the platform split)
- `tests/shared/watcherNativeRecursive.test.js` (new)

No new dependencies. No changes to `src/shared/collector.js`, `src/shared/usage.js`, or any other consumer-facing surface. The event type going from chokidar-shaped to a synthetic `'change'` is the only visible contract change, and it's invisible to the collector (the only event-type consumer is the scheduleTick reason string).

## Out of scope

- Replacing the chokidar stub-based test seam in the in-process host. The in-process host intentionally stays on chokidar to keep the test stub working and to give the worker-fallback path a stable backend.
- Restructuring the watcher to remove the worker-thread isolation. Even with native close resolving synchronously, the worker isolation still helps: a collector replacement recycles the worker to release any native allocation high-water, and the ENOSPC/EMFILE/ENFILE sticky fallback path is unchanged.
- Upstreaming the in-process-host chokidar change. It's a no-op comment update on the existing code path.

---
## Summary by cubic
Uses native macOS `fs.watch({recursive: true})` per top-level watch root in the worker instead of chokidar's per-directory handles, dropping the installed app's memory high-water from ~2.2 GB to ~625 MB total RSS on multi-thousand-dir trees (e.g. OpenClaw's `~/.openclaw/agents/`). The consumer-facing `('all', event, filePath)` event shape is unchanged, so collector logic does not change.

| Area | Before | After |
| --- | --- | --- |
| macOS watch backend | chokidar, one handle per directory | `fs.watch(root, { recursive: true })`, one FSEvents stream per top-level root |
| Memory on large trees | ~2.2 GB compressed/swapped | ~625 MB total RSS (~131 MB main process) |
| Add/unlink event type | chokidar-specific `add`/`unlink` | Synthetic `change` (only consumed in a log string) |
| Watch ignore matching | Applied at watch creation | Applied post-event (fs.watch has no `ignored` option) |
| macOS root-level events with null filename | Dropped (no refresh signal) | Attributed to the root; scans that root's clients |
| Linux/Windows backend | chokidar | chokidar (unchanged) |

**Behavior and design notes**

- Backend selection is macOS + no polling only; `TOKEN_MONITOR_WATCH_POLLING` still forces chokidar in both directions.
- The in-process watcher host stays on chokidar so the collector's `chokidar.watch` test stub keeps working and the worker-failure fallback keeps a well-trodden backend.
- The native backend's `close()` resolves synchronously, so the worker's existing `closeCurrent` await is cheap.
- Watcher construction is deferred to a microtask so setup-time errors (ENOENT, ENOSPC) reach the worker's listener; a per-dir catch forwards a broken root's failure without suppressing events from the others.
- The native ignore matcher follows the chokidar `ignored` contract (truthy = drop the event, falsy = keep); an early build had the polarity inverted and was fixed. Post-event matching is equivalent to chokidar only because every watch policy is prefix-closed, which a new cross-platform test pins — it runs inside a redirected temp home so it never touches the real profile.
- Null-filename root-level events (delivered on macOS) attribute to the root instead of being dropped, since the root matches that root's client attribution.
- Error/ready handling is scoped per watcher instance: a teardown error no longer masks a replacement watcher's failure, `ready` only fires for the live watcher, and a single broken root suppresses `ready` without suppressing events from the other roots.
- `AGENTS.md` documents the platform split and the prefix-closure invariant; the measured backend numbers and tradeoffs live in `nativeWatcher.js`.
- All existing tests pass; new darwin-only real-fs tests cover strategy selection, nested-directory event delivery, post-event ignore filtering (using the same path shapes as the chokidar test), clean `close()`, and continued event delivery after one root fails to open — they wait for `ready` or fail fast on construction error before triggering file changes, and re-touch until the event lands so a busy fseventsd can't drop the first write. A stubbed-`fs.watch` test covers null-filename attribution and runs on every platform.
- No new dependencies; no changes to collector or usage code.

<sup>Written for commit c5bbc2b0aadd365ee93c1e6da8a6a437e60802b5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/621?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Javis603:
Thanks @Cheddaran for tracing the macOS memory growth back to chokidar's per-directory watchers and for measuring it on a real OpenClaw tree. I independently reproduced the handle, memory, and teardown cost, so the underlying problem is real. I also pushed several follow-up commits to this branch while hardening and validating the native watcher path.

While testing both the original implementation and the current head against actual Codex and Claude sessions, I found one blocking behavior difference. When Codex keeps its rollout JSONL file open and continues appending to it, `fs.watch(root, { recursive: true })` on macOS does not reliably deliver the later content changes. Claude refreshed within seconds, while Codex stayed stale until a manual refresh. A focused open-file append test reproduced the same result, but this write pattern is not covered by the current test suite.

Because this regresses the existing live-refresh behavior, I'm not comfortable merging the backend replacement as-is.

The underlying memory diagnosis still stands, so I implemented a narrower OpenClaw-specific pruning approach in #632. It keeps the existing chokidar backend while excluding unrelated runtime and workspace subtrees, and all checks are passing. If convenient, could you test that PR against the same OpenClaw tree and share the before-and-after memory numbers? That would help confirm whether the narrower change resolves the reported 2.2 GB case.

For now, I'll keep this PR open as a reference for further investigation into a native macOS watcher.

Cheddaran:
Thanks for the careful review on this. Closing per the upstream direction — #632 (now merged) addresses the original OpenClaw memory pressure via a narrower chokidar-pruning approach, without the FSEvents open-file-append regression that came up in testing. The conversation here is preserved as a historical reference for why the native FSEvents approach is not the right path on this codebase. Real-tree validation numbers for #632: 1,825 MB → 633 MB total on the same OpenClaw tree that originally measured 2.2 GB.

PR #632, opened 2026-09-07
fix(collector): bound OpenClaw live watcher scope

Javis603:
## Summary

- keep OpenClaw live watching on its documented per-agent usage sources
- prune unrelated workspace and runtime trees before chokidar allocates directory watches
- retain periodic full Tokscale reconciliation for non-standard JSONL locations

## Why

`~/.openclaw/agents` can contain large runtime and workspace trees. Watching the entire directory recursively causes chokidar to allocate directory watches for paths that do not contain usage data.

We also tested the native macOS watcher approach from #621, but repeated appends to an already-open Codex rollout file were not delivered reliably. This change therefore keeps the existing chokidar backend and narrows only the OpenClaw live-watch surface.

The retained live sources are:

- `<agentId>/sessions/**`
- `<agentId>/session-sqlite-import-archive/**`
- `<agentId>/agent/openclaw-agent.sqlite{,-wal,-shm}`
- `<agentId>/agent/codex-home/{sessions,archived_sessions}/**`

Other paths remain covered by the existing periodic full Tokscale scan, so non-standard layouts may refresh later but are not excluded from final accounting.

## Validation

- `node --test tests/shared/collectorLoadGuards.test.js`
  - 93 passed
- `npm run verify`
  - 4109 passed
  - 0 failed
  - 1 skipped
- synthetic OpenClaw tree with 1,000 nested runtime packages, using the polling fallback:
  - before: 2,012 watched directories, approximately 44.5 MB incremental RSS
  - after: 10 watched directories, approximately 1.4 MB incremental RSS

The synthetic benchmark verifies the pruning boundary. Testing against the original reporter's real OpenClaw tree is still needed to confirm the effect on the reported 2.2 GB case.

Related to #621.

---
## Summary by cubic
Bounds OpenClaw live watching to its documented per-agent usage sources, so chokidar no longer traverses large runtime and workspace trees under `~/.openclaw/agents`. Periodic full Tokscale scans still reconcile non-standard JSONL placements.

**Before / After**

| Area | Before | After |
| --- | --- | --- |
| OpenClaw watch scope | Entire `~/.openclaw/agents` recursively | Only `sessions/`, `session-sqlite-import-archive/`, `agent/openclaw-agent.sqlite{,-wal,-shm}`, and `agent/codex-home/{sessions,archived_sessions}` |

**Implementation notes**

- Keeps the chokidar backend; the native macOS watcher from #621 didn't deliver appends reliably.
- Added a new watch-ignore matcher test suite (93 passing); full `npm run verify` passes.
- Synthetic benchmark with 1,000 nested runtime packages: watched directories drop from 2,012 to 10, incremental RSS from ~44.5 MB to ~1.4 MB.
- Still needs verification against the original reporter's real 2.2 GB OpenClaw tree.

<sup>Written for commit 6f81788d59408827d93287fd767d441babee827d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/632?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/shared/collector.js
+++ b/src/shared/collector.js
@@ -2417,6 +2417,17 @@ function clientsForWatchPath(filePath, rootsByClient) {
 // never recurses into an ignored dir (so the runaway poll is gone), yet a
 // newly created state.db-wal is still seen on the next top-level readdir.
 const HERMES_DB_FILES = new Set(['state.db', 'state.db-wal', 'state.db-shm']);
+// OpenClaw keeps each agent's usage sources in a small set of lanes under
+// ~/.openclaw/agents/<agentId>: legacy/published JSONL under sessions/, doctor
+// migration archives beside it, the current per-agent SQLite store, and Codex
+// app-server rollouts under agent/codex-home. The rest of an agent directory is
+// runtime/workspace state and can contain dependency trees large enough to make
+// chokidar allocate thousands of directory watches. Keep the official source
+// lanes live; Tokscale's periodic full scan remains the fallback for a
+// non-standard JSONL placed elsewhere under agents/.
+const OPENCLAW_TRANSCRIPT_DIRS = new Set(['sessions', 'session-sqlite-import-archive']);
+const OPENCLAW_AGENT_DB_WATCH_PATTERN = /^openclaw-agent\.sqlite(?:-(?:wal|shm))?$/;
+const OPENCLAW_CODEX_HOME_DIRS = new Set(['sessions', 'archived_sessions']);
 // OpenCode discovers only direct opencode.db / opencode-<channel>.db files.
 // WAL/SHM are not database inputs to tokscale, but they are the live-write
 // signals that must remain watched so a transaction committed before a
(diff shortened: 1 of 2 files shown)


Cheddaran:
Thanks for putting this together — I tested #632 against the same OpenClaw tree used to validate the original memory growth.

**Setup**
- OpenClaw tree: `~/.openclaw/agents/` with 14,939 dirs and 116,304 files (the same tree from the original 2.2 GB measurement).
- Two builds, identical `.env` and client set, both launched cold and sampled at 30s/60s/90s after the watcher stabilised.

**Results** (RSS sampled via `ps -o rss=`, all four Token Monitor Electron processes summed)

| Build | Main RSS | Total RSS |
|---|---|---|
| Before (`a0692e7ca7`, no fix) | 1,529 MB | 1,825 MB |
| With #632 | 282 MB | 633 MB |
| **Reduction** | **−82%** | **−65%** |

The narrower chokidar-pruning approach reproduces the 2.2 GB → ~0.6 GB win from the original diagnosis without the FSEvents open-file regression that closed #621. The keep-list in the diff (sessions / session-sqlite-import-archive / agent/{openclaw-agent.sqlite, codex-home/{sessions, archived_sessions}}) matches the documented source lanes; runtime and workspace subtrees stop contributing to chokidar's per-directory watch set, which is what drove the 1.5 GB main process.

For the local build I'll plan to drop the FSEvents work in favour of this path once it lands, so #621 will likely stay open only as a reference.

Happy to run any other checks you'd like — e.g. a longer soak, an explicit open-file-append test against Codex rollouts to confirm the regression case, or a side-by-side on a different tree.

Javis603:
Thanks @Cheddaran, this is exactly the real-tree validation I was looking for. On the same 14,939-directory tree, reducing main-process RSS from 1,529 MB to 282 MB and total RSS from 1,825 MB to 633 MB confirms that the bounded OpenClaw watch scope addresses the original memory pressure without replacing the watcher backend.

No further checks are needed for now. Thanks again for running the comparison.

PR #646, opened 2026-09-09
fix(history): preserve daily archives when reads fail

### LINKED ISSUE #636: Daily history archive: a failed read is treated as an empty archive, which can overwrite archived days
### What happens

`readDailyHistoryArchive` falls back to `{}` for **every** failure mode.

`src/shared/dailyHistoryArchive.js:534`

```js
function readDailyHistoryArchive(options = {}) {
  const read = options.readJson || readJson;
  return normalizeDailyHistoryArchive(read(dailyHistoryArchivePath(options), {}));
}
```

And `readJson` (`src/shared/config.js:45`) returns the fallback for any read error *and* any parse error, logging only a warning:

```js
try { content = fs.readFileSync(filePath, 'utf8'); }
catch (error) {
  if (error.code !== 'ENOENT') console.warn(`Could not read ${filePath}: ${error.message}`);
  return fallback;          // EACCES / EBUSY / EPERM land here too
}
if (!content.trim()) return fallback;
try { return JSON.parse(content); }
catch (error) {
  console.warn(`Could not parse JSON in ${filePath}: ${error.message}`);
  return fallback;          // truncated / partially-written JSON lands here
}
```

So a **transient** failure — antivirus holding the file, a partially written file being read mid-write, a lock or permission error on Windows — is indistinguishable from "there is no archive yet".

### Why it matters

`retainDailyHistory` (`dailyHistoryArchive.js:573`) then treats that result as a legitimate empty archive, rebuilds the days from the current in-memory graphs, and writes the result back:

```js
const previous = readDailyHistoryArchive(options);   // {} on failure
let next = capture(previous);                        // rebuilt from graphs only
```

Anything that was in the archive for days no longer present in `graphs` is silently dropped, and the next write persists the truncated version. The caller gets no signal that the baseline was unknown.

I hit this three times on a Windows machine. Each time the archive was momentarily unreadable, the merge re-materialised rows from the raw graphs and duplicated a provider's monthly total — one incident produced 89.3M tokens of duplicated rows. It was recoverable only because I happened to keep a backup.

### Suggested direction

Distinguish "genuinely empty" from "state unknown":

- ENOENT and a blank file are legitimately empty -> keep returning an empty archive.
- Any other read error, and any JSON parse error, should surface as a **failure**, not as an empty archive.
- `retainDailyHistory` should abort the merge (and skip the write) when the baseline is unreadable, rather than writing a half-rebuilt archive.

Concretely: a strict reader returning something like `{ ok: false }`, with `retainDailyHistory` skipping the cycle when it sees that. Skipping one collection cycle is cheap; rebuilding from an unknown baseline is not.

Happy to open a PR with tests (unreadable / truncated / locked-archive cases) if you'd rather have code — I have them locally, just say the word.

### Environment

- token-monitor `main @ e07aeb44`
- Windows 11, Electron desktop build

shuming9315-Fuxiao:
When the daily history archive cannot be read, the existing fallback treats it as empty and can overwrite dates that are no longer present in the current graph. This change gives the archive a strict reader while leaving shared `config.readJson()` behavior unchanged.

Only `ENOENT` initializes an empty archive. Other I/O errors, blank files, malformed JSON, and invalid top-level archive containers propagate through the existing archive error path without writing. Both retention entry points use the strict reader for the initial read and pre-write rebase; the next collection can recover normally. Existing files are preserved without automatic repair.

Validation on Windows / Node v24.16.0:
- `node --test tests/shared/dailyHistoryArchive.test.js tests/shared/collectorHistory.test.js`: 66 passed.
- All 19 added regression cases fail against unchanged main `bda6ffc` (47 existing cases pass), and pass with the fix.
- `npm run verify`: lint passed; 4,140 tests passed, 8 skipped, 1 failed. The same single failure occurs on unchanged main (4,121 passed): `macWidgetLaunchServicesRecovery.test.js` cannot create a symlink on this Windows environment (`EPERM`).
- `git diff --check`: passed.

Tests cover both retention entry points, initial/rebase I/O failures, unchanged archive bytes and timestamps, old-date preservation after recovery, missing-file initialization, and collector error reporting.

Fixes #636

---
## Summary by cubic
Fixes #636 by hardening daily history archive persistence so failed reads cannot overwrite retained dates.

| Area | Before | After |
| --- | --- | --- |
| Archive reads | Any read failure could fall back to an empty archive. | Only a missing file (`ENOENT`) initializes an empty archive; other I/O, blank-file, malformed-JSON, and invalid-container errors propagate without writing. |
| Recovery | A failed read could discard dates absent from the current graph. | Existing archive contents remain untouched, and the next successful collection can recover normally. |

**Design and validation**

- Both retention entry points use the strict reader for initial reads and pre-write rebases, while shared `config.readJson()` behavior remains unchanged.
- Existing invalid files are not repaired automatically.
- Targeted archive and collector tests pass: 66 passed.
- `npm run verify` passed lint with 4,140 tests passed and 8 skipped; the one symlink failure is unchanged on Windows and reports `EPERM`.
- `git diff --check` passed.

<sup>Written for commit 4ef259d7666e2374863b41cef0961bc25ee379e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/646?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/shared/dailyHistoryArchive.js
+++ b/src/shared/dailyHistoryArchive.js
@@ -3,7 +3,7 @@
 const fs = require('node:fs');
 const path = require('node:path');
 const { isDeepStrictEqual } = require('node:util');
-const { readJson, sharedDataDir, writeJsonAtomic } = require('./config');
+const { sharedDataDir, writeJsonAtomic } = require('./config');
 const {
   normalizeTokscaleClientName, num, sumOutputTokens, sumTokens
 } = require('./history');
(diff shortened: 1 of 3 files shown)


Javis603:
Thanks @shuming9315-Fuxiao for tracing the archive overwrite path and covering both retention entry points. The strict reader now fails closed for unreadable or malformed archives without changing the shared config.readJson() behavior, and the pre-write rebase handling closes the remaining data-loss path cleanly.

Looks good to me. Thanks for the thorough regression coverage.

PR #657, opened 2026-09-10
fix(dsh): discover versioned session transcripts so Session Detail finds them

hedanbaomi:
Narrowed to the session-detail half, as you asked — this is now **discovery + Session Detail only**. The usage-plane work (`providers/dsh/usage.js` plus the catalog/collector/partition/WSL wiring) is parked in my fork as branch `feat/dsh-native-usage`, stacked on this branch; I did not open it as a PR against `main` because a stacked base can't live in a fork and I would rather not hand you a duplicate diff while you are deciding. Happy to open it, or to drop it, on request.

## Why the discovery half is still needed after the tokscale fix

`sessionFiles.js` matched only `session.jsonl` / `session.jsonl.zstd`, so for any session the v3 harness re-encoded:

- Session Detail could not open it at all (`findDshSessionFile` never returned a path), and
- the header index that resolves `startedAt`/`lastUsedAt` kept reading the **stale pre-upgrade** copy — the file is still on disk, it just stopped being appended to.

A v3 upgrade leaves both files beside each other, so discovery now matches the version segment and **lists the versioned transcript first**, which is what makes the first-match callers read the file the harness is still writing.

Changes: `sessionFiles.js` (versioned-name match + per-directory ordering, `indexDshSessionHeaders` keeps the first match instead of overwriting) and a behavior-preserving extraction inside `sessionDetail.js` so the record rules live in one function rather than being spread through the loop. `docs/providers/dsh.md` documents the layout, the discovery rules and the plane split.

## On the tokscale side (two things I verified, in case they help your fix)

1. **The records parse fine under the old name.** Copying a v3 transcript to `session.jsonl.zstd` in a scratch `DSH_HOME` reproduced tokscale's exact counts for that session (1,047,701 input / 179,435 output / 485 messages). So the missing name really is the whole bug.
2. **Adding the name should not double-count.** With the same records exposed under two names tokscale already matches, it still reported one session's count (485, not 970) — and with a 2-call copy beside the complete transcript it reported the complete count (485) in both directions. So it unions every matched transcript of a session and dedupes by record signature; a machine keeping both encodings stays correct.

Caveat worth knowing while you're in there: tokscale resolves DSH from the `DSH_HOME` **environment variable**, so `--home <dir>` does not redirect it — a per-home scan (the WSL path) keeps reading the host's `~/.dsh` unless the child gets `DSH_HOME`.

## Tests

`tests/shared/dshSessionFiles.test.js`: versioned names accepted, near-misses rejected (`session.jsonl.lock`, `audit.jsonl`, `*.tmp`), a live versioned transcript ordered before its stale copy, and the header index preferring it. `tests/shared/dshSessionDetail.test.js`: Session Detail opens a session stored under the versioned name.

Full suite on this branch: **4216 tests, 5 failures** — the same five pre-existing environment failures as an untouched `main` checkout here (`clientDataDirPresence` ×3, the packaged-Widget symlink case, the Hermes watch-root case); I checked them against a clean worktree rather than assuming.


Javis603:
Thanks @hedanbaomi for tracing this to the `session.v3.jsonl.zstd` rename and confirming that the existing Tokscale parser can already read the records when the file is exposed under the old name.

The versioned file discovery is still valuable in Token Monitor for Session Detail. Even after Tokscale is fixed, our own session-file lookup would otherwise miss v3 sessions and could keep selecting the stale pre-upgrade transcript. I’d like to keep that part, including preferring the live versioned file when both copies exist.

For usage collection, though, I’d prefer not to move DSH to `locallyParsed` or duplicate Tokscale’s parsing, pricing, deduplication, history, and WSL responsibilities here. The missing filename should be fixed in Tokscale. If an upstream release is not available soon enough, I can temporarily pin a build from my Tokscale fork.

Would you be open to narrowing this PR to the Session Detail and versioned file-discovery changes and their tests, and dropping `providers/dsh/usage.js` together with the catalog, collector, partition, and WSL usage wiring?

hedanbaomi:
Both verified while narrowing, in case they save you time on the tokscale side:

**1. The records parse correctly under the old name.** A scratch `DSH_HOME` containing a v3 transcript copied to `session.jsonl.zstd` reproduces tokscale's exact counts for that session — 1,047,701 input / 179,435 output / 485 messages. So the versioned name is the whole gap, not the record format.

**2. Adding the name to the matcher should be safe.** I checked the two ways it could bite:

| Setup | Reported |
| --- | --- |
| One transcript | 485 messages (baseline) |
| Same records under two names tokscale already matches | **485** (not 970) |
| 2-call copy beside the complete transcript, and the reverse | **485** both directions |

So tokscale unions every matched transcript of a session and dedupes by record signature: a home that kept the pre-upgrade file beside the re-encoded one stays at one count, and it does not matter which of the two it reads first. That removes the only reason I could see for the widget to own a second DSH usage path, which is why I am happy to leave that half parked in the fork (`feat/dsh-native-usage`, stacked on this branch) and to drop it entirely once a release carries the fix.

One caveat from the same investigation: tokscale resolves DSH from the `DSH_HOME` environment variable, so `--home <dir>` does not redirect it. The per-home scans (WSL) therefore keep reading the host's `~/.dsh`; if that path is meant to cover a distro's DSH sessions it will need `DSH_HOME` set on the child instead of `--home`.

Javis603:
Thanks @hedanbaomi for tracing the versioned DSH transcript layout, validating the stale-copy behavior, and narrowing this PR to discovery and Session Detail.

I kept generic `session.vN...` discovery as a best-effort path, then pushed follow-ups for live generation promotion, inherited v3 seed boundaries, and attempt-only usage accounting. The matching Tokscale build is pinned too, so aggregate usage and Session Detail now follow the same accounting.

hedanbaomi:
thanks for the review

hedanbaomi:
**Correction to my earlier comment: the WSL / `--home` caveat I posted there was wrong.**

I wrote that tokscale resolves the DSH root only from the `DSH_HOME` environment variable, so `--home <dir>` cannot redirect it and the per-home (WSL) scans read the host's `~/.dsh`. That is not what happens, and I should not have published it without verifying it properly. Retested with unique probe ids, so which transcript gets read is unambiguous:

| Setup | Session read | Conclusion |
| --- | --- | --- |
| `--home T`, `DSH_HOME` unset (probe at `T/.dsh/sessions/...`) | `probe-via-home-flag` | `--home` **does** redirect DSH |
| `--home <empty dir>`, `DSH_HOME` unset | none | no fallback to the host's `~/.dsh` |
| no `--home`, `DSH_HOME=X` | `probe-via-env-var` | the env var is used when there is no `--home` |
| `--home T` **and** `DSH_HOME=X` | `probe-via-home-flag` | `--home` wins; `DSH_HOME` is ignored |

So the DSH root is `<--home>/.dsh` when `--home` is given (env roots disabled — the `use_env_roots: false` path), and `$DSH_HOME` → `~/.dsh` only without `--home`. Per-home WSL scanning therefore already covers each distro's `~/.dsh/sessions`, and a host-side `DSH_HOME` cannot leak into it. **Nothing needs changing on that path** — please disregard that paragraph.

How I got it wrong: my first probe put the transcript at `<tmp>/sessions/...`, missing the `.dsh` segment, so `--home <tmp>` correctly found nothing and I read that as "`--home` is ignored", then repeated it without re-checking. The rest of that comment still stands — the dedupe finding (a session holding both encodings is counted once; a partial copy beside the complete one does not change the count) came from a separate experiment with its own controls.

Issue #659, opened 2026-09-10
Factory/Droid support request


ceciiialee:
Can you please add support for Droid token collection and Factory plan limits?

REMIXPLAYER:
Thanks for the request — I've investigated both parts:

**Droid token collection: ready to add.** [tokscale](https://github.com/junhoyeo/tokscale), the scanner our token counting delegates to, has parsed Droid since v1.0.18 (Dec 2025) via the cumulative usage in `~/.factory/sessions/**/*.settings.json`. Wiring it in is catalog/roots/icons/docs work on our side — no scanner changes needed.

**Factory plan limits: feasible, but a scope decision.** `GET https://api.factory.ai/api/billing/limits` with a `fk-` API key ([created here](https://app.factory.ai/settings/api-keys)) returns the 5-hour / weekly / monthly window percentages with reset times plus the prepaid credit balance — everything a live quota card needs. Caveat: this is the undocumented endpoint [CodexBar](https://github.com/steipete/codexbar/blob/main/docs/factory.md) also uses, and it has drifted once before; Factory's documented Analytics API can't substitute (day-lagged, org-role-gated). So the limits side carries ongoing maintenance risk that the token side doesn't.

@Javis603 how would you like to scope this — land both (Droid tracking + Factory limits), or start conservative with Droid as a tracked client only and treat the undocumented-billing-endpoint limits provider as a separate follow-up? Happy to send a PR either way.

Javis603:
@REMIXPLAYER Sounds good — let's start with Droid token collection first.

For Factory plan limits, I'd prefer to keep that as a separate follow-up.

REMIXPLAYER:
Following the scope discussion above, the token-collection half is up for review: #682

Droid is in as a default-tracked client — usage through tokscale's existing droid parser, session titles/times/projects from the kernel's `sessions-index.json`, official Factory artwork. Factory plan limits will follow as a separate PR once that design settles.

PR #682, opened 2026-09-12
feat(droid): track Droid CLI and Factory desktop token usage

REMIXPLAYER:
## Summary

- add Factory Droid, covering the Droid CLI and Factory desktop app, as one default-tracked client: one droid kernel, one `~/.factory/sessions` data plane, and zero scanner changes — tokscale has parsed Droid since 1.0.18.
- usage, watch, health, WSL discovery, custom scan paths, icons and docs go through the existing per-client wiring; a session-metadata resolver adds titles, timestamps and project attribution from the kernel's current and legacy session indexes.
- token collection only — Factory plan limits follow separately, per the issue discussion.

## What changes

- `droid` joins the catalog and the scan path: a home-relative `~/.factory/sessions` root under the `droid-sessions` check id, the WSL marker, and the renderer/Discord/docs surfaces.
- one client covers both front-ends: the Factory desktop app bundles the same droid kernel and writes the same sessions tree — no desktop-specific root.
- the user-facing label is `Factory Droid`, positioned after Cline, while `droid` remains the canonical usage and tokscale client id.
- a session-metadata resolver reads the current `.factory/cache/session-discovery-index.json` v6 object map and the legacy `.factory/sessions-index.json` v2 array for titles, timestamps and project identity; current metadata wins when both contain the same session, and `cwd` goes through the shared `projectIdentity()` so droid joins project attribution with the same hashing as every other client.
- after #676 the scan carries droid's `sessions` timestamps; the resolver reconciles (earliest start, latest last-activity) and still owns title and project attribution — the scan carries no `title` and its `workspaces` array is not decoded for droid.
- Droid's disjoint `thinkingTokens` bucket is folded into public output and reconstructed history totals in both the Node and Worker implementations.
- exact-match client-id normalization: the substring form folded `android` into `droid` (`'android'.includes('droid')`).
- icons use the official Factory artwork: the docs.factory.ai favicon vector for the row mask and the Factory.app icon for the README table.
- README tables and counts across all five locales, `.env.example` CSV, chart colors and vendor labels.

## Compatibility

- droid is default-tracked for fresh installs and installs without an explicit saved client selection; a saved explicit `clients` CSV wins verbatim and does not silently grow — existing users enable Droid with one toggle in settings.
- a missing, malformed or entries-less current or legacy session index — or an older droid build that never writes one — resolves to no metadata; session rows keep working with bare ids.
- the current v6 index uses `id`, `createdTimeMs` and `modifiedTimeMs`; the legacy v2 index uses `sessionId` and `mtime` and does not contain `createdAt`, so it supplies `lastUsedAt` without inventing a start time.
- the scan-backed `sessions` array is absent from older tokscale builds; the resolver still supplies whatever title, project and timestamp metadata is available from the indexes.
- BYOK sessions record full local token counts (verified live); subscription accounts are expected to do the same — droid's own `/cost` panel approximates credits from these tokens — but that could not be verified without a subscription.
- `FACTORY_HOME_OVERRIDE` is deliberately not consulted. Per-tool custom scan paths relocate token scanning through `TOKSCALE_EXTRA_DIRS`, but session-index metadata remains home-relative, so relocated sessions may not receive title or project enrichment.

## Validation

- `npm run verify` — 4,329 passed / 0 failed / 1 skipped
- current session-index behavior verified against data generated by Droid 0.218.1
- tokscale JSON smoke test on real Droid sessions: 16,033 input / 42 output from an OpenRouter BYOK session; pricing probe covers a Factory router model, an Anthropic id and a free model ($0)
- live widget check: dashboard row, session row with title and time, project attribution on Home
- icon provenance: row-mask path byte-identical to the docs.factory.ai favicon; README PNG sha-identical to a re-render of the Factory.app icon

The tracked client id is `droid` — tokscale's and the product's own name; a future limits provider registers as `factory` and must not rescan the same sessions tree.

Refs #659 (Factory plan limits will follow separately, per the issue discussion)


REMIXPLAYER:
Factory is the desktop/vendor-facing name, while Droid is the CLI/agent/kernel and the Tokscale client id. The desktop app and CLI write the same ~/.factory/sessions data, so Token Monitor currently tracks the usage side as droid.

Could you confirm the intended project convention: should we keep the internal usage client id as droid, present it to users as Droid (Factory), and reserve factory for a future limits/billing provider rather than adding a second factory usage scanner?

The Droid/Factory naming split is easy to misunderstand, so I would like to align the UI and documentation with your preferred boundary before expanding this work.

Javis603:
Thanks @REMIXPLAYER for spelling out the naming split. Yes, that is the boundary I want: keep `droid` as the canonical usage and Tokscale client id, use `Factory Droid` as the user-facing name, and reserve `factory` for a future limits or billing provider. Since Droid CLI and Factory desktop share `~/.factory/sessions`, they should remain one usage scanner.

I pushed a follow-up commit (ff7290e9) to apply that naming consistently, move Factory Droid after Cline in the canonical tool order, replace the README icon, and keep the related presentation lists aligned. It also folds Droid’s disjoint `thinkingTokens` into public output and history totals, with regression coverage.

Javis603:
One more follow-up after validating the session metadata path against real data from the current Droid 0.218.1 release.

Current Droid uses `~/.factory/cache/session-discovery-index.json` with a v6 object-map schema (`id`, `createdTimeMs`, and `modifiedTimeMs`), while the legacy `~/.factory/sessions-index.json` uses v2 array entries and does not contain `createdAt`.

I pushed (a5e89d0c) to support both schemas, prefer the current index when both exist, and remove the invalid `createdAt` assumption. Tokscale remains the source of token accounting. The Droid session index is used only to supplement metadata that its current Tokscale parser does not expose, including titles and cwd-based project attribution.

REMIXPLAYER:
Much appreciated for the merge and for closing the gaps I couldn't see from an older install — the 0.218.1 index schema and the naming/thinkingTokens follow-through. Happy to pick up the Factory limits follow-up next.

PR #685, opened 2026-09-13
feat(factory): add Factory Droid plan limits via FACTORY_API_KEY

### LINKED ISSUE #659: Factory/Droid support request

REMIXPLAYER:
## Summary

- add a `factory` limits provider for Factory Droid plan limits — the second half of #659 (token collection landed in #682). The probe is a faithful port of CodexBar's working `FactoryStatusProbe`, cross-checked field-by-field against Factory's own CLI binary and web bundle, so every endpoint and response field is either first-hand verified or carried by a production implementation.
- authenticated solely by `FACTORY_API_KEY` (explicitly saved in Token Monitor, the process environment, or Droid's own `~/.factory/.env`). Auto-discovered keys stay in memory, are never persisted, and never reach the hub; Droid's loginkeychain/OAuth credential material is never read (#586 boundary).
- naming per the #682 discussion: `droid` stays the canonical usage client, `factory` is the limits provider id, `Factory Droid` is the user-facing label.

## What changes

- current billing model (token-rate limits): Standard 5-hour / weekly / monthly windows plus optional Core windows and the prepaid extra-usage balance (`extraUsageBalanceCents` → a `credits` window). The response gate follows Factory's explicit `usesTokenRateLimitsBilling` flag and requires `limits.standard`; false or missing flags fall back to the legacy route even when rollout fields are already present.
- legacy billing model: `/api/organization/subscription/usage` fallback with a hardened percent heuristic against stale ratios and absurd allowance values.
- settings UI: an account panel with an `fk-` input and not-configured / from-.env / configured / invalid states; limits presentation, tray and widget identity, i18n across five locales.

## Compatibility

- fresh installs seed `factory` when Droid is detected (the `droid → factory` bridge in `LIMIT_PROVIDER_BY_CLIENT`); existing explicit provider selections keep user intent — enable Factory Droid with one toggle.
- a missing or malformed `~/.factory/.env` degrades to not-configured; nothing else changes for other providers.

## Validation

- `npm run verify` — 4,369 passed / 0 failed / 2 skipped
- per-request capture proves the settings > env > `.env` credential precedence, that a discovered key is used in-memory only, and that no key means zero network requests
- a dedicated test pins that an existing `auth.v2.loginkeychain` never satisfies configuration
- response shapes cross-checked against three sources: Factory's own CLI binary and web bundle (first-hand), CodexBar's working probe (second-hand), and the official pricing model (5-hour / 7-day / 30-day windows + prepaid Extra Usage)

## Screenshots

No real Factory subscription was available for testing — the quota windows below are rendered from injected sample data built on the reverse-engineered response structure (fields verified first-hand in Droid's CLI and web bundle).

| | |
| --- | --- |
| **The `factory` limits account panel** — not-configured, before first key | |
|  | |
| **Quota windows** — injected demo, real windows need an `fk-` key | |
|  | |
| **The `droid` usage client** (from #682) | |
|  | |

## Scope

Closes #659 — token collection landed in #682 and this PR completes the limits half. Per the #682 naming discussion and the #586 credential boundary: `droid` remains the canonical usage client, this provider registers as `factory` with `Factory Droid` as the user-facing label, credentials are API-key only, and the provider and usage documentation share one origin in `docs/providers/droid.md`.


Javis603:
Thanks @REMIXPLAYER for taking the Factory limits half of #659 through the API mapping, credential boundary, account UI, documentation, and tests.

I pushed two follow-up commits before merging. 80b10d00 preserves widget-owned Factory quota across headless-agent updates and keeps invalid saved keys recoverable. 4898508d follows Factory's explicit billing-mode flag, links to the API keys page, and matches CodexBar's production request shape.

REMIXPLAYER:
Thanks @Javis603 for the careful review and follow-up fixes. I really appreciate you catching the lifecycle and edge-case gaps I missed — very helpful for improving how I approach these integrations.🥰

PR #606, opened 2026-09-04
feat(unsloth): add Unsloth Studio usage tracking

The-Minion-oOo:
Adds Unsloth as a regular tracked client, following the [scope clarification in #596](https://github.com/Javis603/token-monitor/pull/596#issuecomment-5535993994).

The current Tokscale 4.15.1 dependency already includes the parser and provider-pricing follow-up, so this does not change dependencies or add another parser.

- Detects `studio.db` under `UNSLOTH_STUDIO_HOME` or `~/.unsloth/studio` and watches only the database and its WAL/SHM sidecars.
- Adds the client to settings, tool rows, chart colors, tray artwork, source health, and WSL discovery. Existing saved tool selections are preserved.
- Updates all five READMEs and documents the inference-only and pricing boundaries. Local inference keeps zero API cost; recognized metered providers use Tokscale estimates.

Discord uses the Unsloth text label for now. The application asset list does not contain `unsloth`; the supplied PNG can be uploaded by the app owner before enabling its small-image key.

## Validation

- `npm run verify`: lint passed; 4,002 tests passed, 8 skipped, 4 failed. The same four failures reproduce on untouched v0.53.0 on this Windows machine: the macOS widget symlink fixture, Antigravity source presence, Copilot source presence, and Hermes watch pruning.
- `node scripts/verify-vendored-tokscale-clients.js`: passed against the installed Windows Tokscale 4.15.1 binary.
- Windows CI exposed a short-path mismatch in the new test fixture. Using a canonical temp root fixes it; all seven Unsloth tests also pass with an aliased temp directory.
- WAL matching and targeted client attribution are checked without depending on OS event timing; the shared collector suite covers event delivery.
- A synthetic SQLite fixture against that binary returned 167 local chat/API tokens at $0, then 307 tokens with a positive metered estimate after a WAL append. Private-field sentinels were absent from the output.
- Rendered the actual Electron tool rows in dark and light themes and checked the searchable Unsloth settings entry with isolated fixture data. No installed settings or live conversation data were changed.

Linux/macOS runtime behavior, live WSL database access, and live Discord presence were not tested on this Windows machine.

## Screenshots

Tool rows with fixture data, in dark and light themes:

| Dark | Light |
|:---:|:---:|
| Unsloth usage in dark theme | Unsloth usage in light theme |

PR #694, opened 2026-09-15
feat(amp): track Amp (AmpCode) token usage

Javis603:
Adds Amp (AmpCode) as a tracked usage client. Tokscale's `amp` client already parses Amp's thread files, so this wires the existing support into Token Monitor instead of adding a local adapter.

## Data source

Amp writes one JSON thread per session (`T-*.json`) under the XDG data root. Tokscale declares this as `PathRoot::XdgData` + `amp/threads` in `clients.rs`, so the watcher and health check derive the root from `$XDG_DATA_HOME` (`~/.local/share/amp/threads` by default) rather than a home-relative literal — a relocated data home moves the watcher with it. Every thread holds a `usageLedger` plus per-assistant-message `usage`, which Tokscale reconciles into input / output / cache-read / cache-write.

Verified against the vendored binary rather than the manifest alone:

```console
$ tokscale clients
  Amp
  sessions: ~/.local/share/amp/threads ✓
  messages: 1
```

## Changes

- **Identity** — `amp` / "Amp" in `CLIENT_CATALOG`, placed after Cline and before Factory Droid (the upstream order that keeps the new-install list, settings page and README table in one sequence). Default-tracked; not a parse-local client.
- **Collection** — the XDG-derived `amp/threads` source root, the `amp-threads` health check id, a WSL marker plus its `MARKER_CLIENTS` attribution, and `normalizeClientName` matching `amp` exactly (an exact match keeps the id a fixed point for targeted watch ticks without swallowing a differently-named client containing the same substring).
- **Presentation** — `.row-icon-amp` mask rule, `assets/icons/amp.svg`, `clientsWithIcon`, `VENDOR_ORDER` / `VENDOR_LABELS`, the Discord asset/label maps, and the macOS Widget vendor colour, all keyed on Amp's brand red `#F34E3F`.
- **Docs** — the `.env.example` client CSV and all five README tables, with prose counts corrected to 37+/30+ so `readmeConsistency` stays exact.
- **Hub** — `usage.js` and `clientHealth.js` sit in the portable Hub core, so `hubBuildRegistry.json` was regenerated and `npm run sync:worker` re-run.
- **XDG environment reconciliation** — blank subprocess values, the effective Windows home fallback, and caller-injected environments are kept in lockstep with the spawned scan; see below.

## Icon and colour

The row icon is the official LobeHub Amp glyph (the same MIT-licensed set already used for Unsloth, attribution header kept), carrying the brand red `#F34E3F` rather than the `currentColor` most row marks use. Amp's own published asset is a wordmark, which is illegible at the 10px row size, so the glyph is the mark that actually reads in the list.

Two consequences of the fixed fill are deliberate and covered by tests:

- **The row list renders the mask, not the fill.** `.row-icon-amp` paints `currentColor` through `mask-image`, so the row mark stays theme-tinted and the SVG's fill does not leak into the list.
- **The tray re-inks it.** The glyph is a single flat ink, so `isFlatInkPixels` classifies it as re-colourable and the tray paints it in the menu-bar ink instead of the brand red.

The README table icon is a 640px asset matching the other tool icons' dimensions.

## XDG environment reconciliation

Review found that a blank `XDG_DATA_HOME` made the watcher and the scan disagree. Confirmed against the vendored binary:

```console
$ XDG_DATA_HOME unset -> scan finds ~/.local/share/amp/threads (1 message)
$ XDG_DATA_HOME="   " -> scan resolves "   /amp/threads", 0 messages
```

Tokscale reads these roots with a bare `std::env::var(...)`, so ANY present value wins; Token Monitor resolves the same roots with `nonBlankEnvPath()`, which treats a blank value as unset (matching the XDG basedir spec: `$XDG_DATA_HOME` is "either not set or empty"). The result was a `detected` health state beside a scan of a different, non-absolute path.

This is **pre-existing and wider than Amp**: the same read covers `XDG_CONFIG_HOME` and `TOKSCALE_HEADLESS_DIR`, and through `PathRoot::XdgData` / `PathRoot::Config` it reaches opencode, kilo, crush, goose, zed, micode, devin-cli, hindsight, antigravity, trae, warp and mcode. A blank value is also easy to hit by accident: `loadDotEnv()` copies a bare `XDG_DATA_HOME=` line from `.env` into `process.env` as `""`.

The fix is scoped to Token Monitor: `tokscaleCommand()` — the single outlet every tokscale subprocess goes through — drops those three keys from the subprocess environment when their value is blank, so tokscale takes its own fallback, which is the root Token Monitor already resolved. Both sides then agree, and the behaviour stays correct if tokscale later adopts blank-as-unset itself. No upstream change or vendored rebuild is involved. `TOKSCALE_CONFIG_DIR` is deliberately excluded because both sides already agree: an empty value is unset, while any non-empty value (including whitespace-only) is treated as an override.

The source-root side now also derives the XDG fallback from `tokscaleHomeDir({ env, platform, homeDir })` and passes the `env` injected into `clientSourceRoots()` through to `xdgDataHome()`. This keeps watcher and health roots aligned with the scan when Windows uses an absolute native `HOME` override, and prevents a caller-provided `XDG_DATA_HOME` from being replaced by `process.env`.

One case is knowingly left alone and documented in the comment: a *non-blank but relative* `XDG_CONFIG_HOME`, which Token Monitor rejects via `absoluteEnvPath()` while a raw read would accept it. That is a separate divergence on a value the spec forbids, and it lives in a Linux-only arm this repo's test matrix cannot exercise.

## Deliberate non-changes

These surfaces are touched for other tracked clients but correctly do not apply here, verified against Tokscale's parser:

- **No session-metadata resolver.** `AmpThread` has no title field, so there is nothing for `sessionMetadata.js` to read.
- **Not in the disjoint-reasoning set.** Amp's records are written with `reasoning: 0`, so folding a reasoning bucket would add nothing.
- **No limits provider.** Amp is usage-only in this repo; no `LIMIT_PROVIDER_CATALOG` entry or collector fetcher.
- **No tray `SPECIAL_ICON_SOURCES` entry.** The tray resolver already falls back to `assets/icons/<id>.svg`.

## Validation

New `tests/shared/amp.test.js` covers id normalization (including that unrelated names containing "amp" are not absorbed), the XDG root plus its blank-value behaviour, watcher/health paths, custom scan roots, WSL discovery, token extraction, and that workspace-less threads stay unattributed instead of being labelled from the thread id. New `tests/shared/tokscaleBlankEnv.test.js` pins the reconciliation directly: a blank key is dropped, a real override is kept, the caller's object is not mutated, `TOKSCALE_EXTRA_DIRS` survives, and the returned object is identity-equal when nothing needs dropping. It also covers case-insensitive Windows environment keys, Tokscale's effective Windows `HOME` fallback, and caller-injected environment precedence over `process.env`. The existing XDG guard test now includes Amp alongside opencode/zed/micode, and a presentation test asserts the brand colour, the mask rule and that the mark stays gradient- and stroke-free for the tray's re-ink path.

End-to-end against the vendored binary, through `tokscaleCommand().env`: blank, whitespace-only, all-three-blank and unset all produce the same result (the scan finds the thread, the watcher resolves `~/.local/share/amp/threads`), while a real absolute override still reaches the subprocess.

- `tests/shared/amp.test.js` — 7 passing; `tests/shared/tokscaleBlankEnv.test.js` — 9 passing
- Focused Amp, source-root, health and Tokscale suites — 183 passing
- Registration, catalog, health, presentation, docs and Hub-build guards — passing
- ESLint over `src` and `tests` — clean
- `git diff --check` — clean

Note on the local environment: the full suite in this worktree shows pre-existing failures that reproduce unchanged on `main` (sandbox `listen EPERM`, `EMFILE` on native file watchers, SQLite lock contention) plus this worktree having no `node_modules` of its own. None of them are related to this change.

---
## Summary by cubic
Adds Amp (AmpCode) as a default-tracked Tokscale client, reusing Tokscale’s parser for `T-*.json` thread files instead of a local adapter. It also makes XDG root resolution and subprocess environment handling consistent, so scans, watchers, and health checks no longer disagree about the source tree.

| Area | Before | After |
| --- | --- | --- |
| Amp usage | Not tracked | Tracked by default from `$XDG_DATA_HOME/amp/threads` (`~/.local/share/amp/threads` by default) |
| Blank XDG values | Token Monitor treated blank values as unset while Tokscale received them and could scan a different root | Blank `XDG_DATA_HOME`, `XDG_CONFIG_HOME`, and `TOKSCALE_HEADLESS_DIR` values are removed before launching Tokscale |
| Windows and injected environments | Watcher and health paths could use a different home or `process.env` value than the scan | Roots follow Tokscale’s effective Windows `HOME` and the caller-provided `env` |

**New Features**

- Registers `amp` with exact-name normalization, custom scan roots, WSL discovery, and an `amp-threads` health check.
- Adds Amp’s MIT-licensed icon and brand red `#F34E3F` to rows, charts, tray re-inking, macOS Widget colours, Discord assets, and documentation.
- Amp threads have no workspace or title, so sessions remain unattributed and receive no session metadata, reasoning, or limits support.
- Blank-key matching is case-insensitive on Windows and exact on POSIX; non-blank values and the caller’s environment object remain unchanged.

**Validation**

- Added coverage for XDG roots, blank handling, effective Windows `HOME`, injected `env` precedence, WSL discovery, token extraction, unattributed sessions, and presentation assets.
- Collector, usage, Tokscale, catalog, health, documentation, Hub-build, ESLint, and diff checks pass; unrelated full-suite sandbox, native watcher, and SQLite lock failures remain reproducible on `main`.

<details>
<summary>中文版本</summary>

新增 Amp（AmpCode）作為預設追蹤的 Tokscale 用戶端，重用 Tokscale 現有的 `T-*.json` 執行緒檔案剖析器，而不是加入本機轉接器。此變更亦統一 XDG 根目錄解析及子程序環境處理方式，令掃描、監看及健康狀態檢查不再使用不同的來源目錄樹。

| Area | Before | After |
| --- | --- | --- |
| Amp 用量 | 未有追蹤 | 預設從 `$XDG_DATA_HOME/amp/threads` 追蹤（預設為 `~/.local/share/amp/threads`） |
| 空白的 XDG 值 | Token Monitor 將空白值視為未設定，但 Tokscale 仍會接收這些值，可能掃描另一個根目錄 | 啟動 Tokscale 前會移除空白的 `XDG_DATA_HOME`、`XDG_CONFIG_HOME` 及 `TOKSCALE_HEADLESS_DIR` 值 |
| Windows 及注入的環境 | 監看及健康狀態路徑可能使用與掃描不同的主目錄或 `process.env` 值 | 根目錄跟隨 Tokscale 在 Windows 上的有效 `HOME` 及呼叫者提供的 `env` |

**新功能**

- 將 `amp` 註冊並加入完全相符的名稱正規化、自訂掃描根目錄、WSL 偵測及 `amp-threads` 健康狀態檢查。
- 將 Amp 的 MIT 授權圖示及品牌紅 `#F34E3F` 加入列、圖表、托盤重新著色、macOS Widget 顏色、Discord 素材及文件。
- Amp 執行緒沒有工作區或標題，因此工作階段會保持未歸屬，亦不會加入工作階段中繼資料、推理分類或額度支援。
- 空白金鑰的比對在 Windows 上不區分大小寫，在 POSIX 上則完全比對；非空白值及呼叫者的環境物件均保持不變。

**驗證**

- 新增 XDG 根目錄、空白值處理、Windows 有效 `HOME`、注入 `env` 的優先次序、WSL 偵測、Token 擷取、未歸屬工作階段及呈現素材的測試覆蓋。
- Collector、usage、Tokscale、目錄、健康狀態、文件、Hub 建置、ESLint 及 diff 檢查均通過；完整測試套件中與本變更無關的 sandbox、原生檔案監看器及 SQLite 鎖定錯誤，仍可在 `main` 重現。

</details>

<sup>Written for commit 1ac34f49d24e3b75e050f1bdf89694de3979c6b6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/694?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Javis603:
Thanks — the finding is real, and I confirmed it against the vendored binary rather than by reading alone:

```console
$ XDG_DATA_HOME unset -> scan finds ~/.local/share/amp/threads, 1 message
$ XDG_DATA_HOME="   " -> scan resolves "   /amp/threads", 0 messages
$ XDG_DATA_HOME=""    -> same, 0 messages
```

while Token Monitor resolved `~/.local/share/amp/threads` for the watcher and health check on those same inputs — exactly the "health says detected, scan reads somewhere else" case you described. It also holds structurally: `tokscaleCommand()` inherits `process.env` and passes no `--home`, and the CLI gates on `use_env_roots = opts.home_dir.is_none()`, so env roots are on for normal scans.

**One correction to the scope.** This is pre-existing and wider than Amp. The same raw read covers `XDG_CONFIG_HOME` and `TOKSCALE_HEADLESS_DIR`, and through `PathRoot::XdgData` / `PathRoot::Config` it reaches opencode, kilo, crush, goose, zed, micode, devin-cli, hindsight, antigravity, trae, warp and mcode. Amp only made it reachable for a client we newly ship. Upstream is also internally inconsistent here — `CLINE_SESSION_DATA_DIR`/`CLINE_DATA_DIR`/`CLINE_DIR` go through `non_blank_env_path` and the `$XDG_DATA_HOME/gjc` branch trims explicitly, so the bare `var(...)` reads look like an omission rather than intent. Worth noting the trigger is easy to hit: `loadDotEnv()` copies a bare `XDG_DATA_HOME=` line from `.env` into `process.env` as `""`.

**Fix, scoped to Token Monitor as agreed.** Rather than teaching the shared resolvers to reproduce the current binary (which would bind nine other XDG clients to the same behaviour), `tokscaleCommand()` — the single outlet every tokscale subprocess goes through — now drops those three keys when their value is blank, so tokscale takes its own fallback and both sides land on the root Token Monitor already resolved. Blank and unset then behave identically, and it stays correct if tokscale adopts blank-as-unset itself, so no upstream change or vendored rebuild is involved. `TOKSCALE_CONFIG_DIR` is deliberately excluded since both sides already treat a blank value as set.

End-to-end through `tokscaleCommand().env` against the vendored binary: blank, whitespace-only, all-three-blank and unset now all scan the thread and resolve `~/.local/share/amp/threads`, while a real absolute override still reaches the subprocess.

The test comment you flagged is rewritten — it no longer presents the ambiguous case as Amp's contract, and it records the divergence and the affected clients instead.

One case is knowingly left alone and documented in the code comment: a *non-blank but relative* `XDG_CONFIG_HOME`, which Token Monitor rejects via `absoluteEnvPath()` while a raw read would accept it. That is a separate divergence on a value the spec forbids, and it lives in a Linux-only arm this repo's test matrix cannot exercise, so I did not fold it into this change.

Javis603:
Both points handled in `1072da16`.

## Windows env-key casing

Confirmed the bug before fixing it — a mixed-case key really did survive the drop:

```console
in : { Path: 'C:\\Windows', Xdg_Data_Home: '   ', HOME: '/home/u' }
out: { Path: 'C:\\Windows', Xdg_Data_Home: '   ', HOME: '/home/u' }   # still blank
```

because `{ ...process.env }` preserves the casing Node was handed, so `env['XDG_DATA_HOME']` was `undefined` on the plain object and the `typeof` guard skipped the key entirely.

`tokscaleEnvWithBlanksDropped(env, platform)` now matches case-insensitively on `win32` and keeps the exact-match rule on POSIX, where names are case-sensitive and `Xdg_Data_Home` is a genuinely different variable that must not be touched. `tokscaleCommand()` passes `options.platform || process.platform`.

Two tests cover it, so the CI matrix is not the safety net: a win32 case that drops `Xdg_Data_Home` while keeping the mixed-case `Path` every Windows shell sets (and keeps a non-blank `Xdg_Data_Home`), and a POSIX case asserting the mixed-case key is left completely alone.

Verified: the object above on win32 leaves only `Path`; the same object on linux is returned untouched.

## TOKSCALE_CONFIG_DIR wording

You are right, and it was imprecise in both directions — it said "a blank value counts as set", but an empty string is unset while whitespace is set. Now reads:

> `TOKSCALE_CONFIG_DIR` is deliberately NOT here, because both sides already agree on it: an empty value is unset, while any non-empty value — whitespace included — is an override. Tokscale spells that `!custom.is_empty()` and Token Monitor `override.length > 0`, so there is nothing to reconcile.

## Amp test comment

The stale "the fix belongs in tokscale" framing is gone. It now points at the reconciliation that actually shipped (`tokscaleEnvWithBlanksDropped` and `tests/shared/tokscaleBlankEnv.test.js`) and explains why the scan and the assertion agree, rather than describing an upstream fix this PR no longer waits on.

## On relative XDG_CONFIG_HOME

Agreed — leaving it as the documented comment, follow-up issue if you want it tracked. It is invalid per spec (`$XDG_CONFIG_HOME` must be an absolute path), so it is upstream's treatment of invalid input rather than anything Amp introduced.

Javis603:
Thanks — the collection/aggregation/health findings all check out and the two comment nits are fixed in `1072da16`. On the Discord asset I traced the history before acting, and I think the premise needs correcting rather than the code changing.

**`KNOWN_CLIENT_ASSETS` has a hard CI guard, so omitting `amp` is not an available option.** #551 added it, and `tests/shared/clientRegistrationConsistency.test.js` asserts `missingFromAssets` is empty for every `KNOWN_CLIENTS` id:

```
with amp    -> missingFromAssets = []        -> guard passes
without amp -> missingFromAssets = ["amp"]   -> guard FAILS
```

The assertion message is itself explicit that an entry is not a promise the asset exists: *"Rich Presence will show no icon for these clients when they are the top client today."* So the set means "keys we attempt to send", and the failure mode of an un-uploaded asset is a missing small image, not a broken payload.

**The two precedents read the other way.** I checked both commits:

- **#606 (Unsloth)** — `4a280443` **does** add `unsloth` to `KNOWN_CLIENT_ASSETS` in the same commit that adds the client. Its docs note says the small image *"can be enabled … after the app owner uploads an asset"*, but the code entry shipped with the client regardless. The note is forward-looking prose, not the deferred-registration practice.
- **#301 / `b89dce73` (Qoder CN)** — this one was the opposite direction entirely. `qodercn` was added as a tracked client **without** the asset entry, which the #551 guard then caught, and `b89dce73` is the fix that **adds** it: `fix(discord): show a Rich Presence icon for Qoder CN`. The commit message says the missing entry made Rich Presence "render the label with no icon", and it was treated as a bug to fix — not a state to preserve.

So the established pattern is: ship the client with the key registered, and treat an absent key as a defect. Deferring would mean landing `amp` with a deliberately red guard test.

**On the external state.** You are right that whether `amp` is uploaded to application `1507034330436862062` is not something the repo can verify, and I cannot check it from here. What I can say is that this PR does not claim the asset is registered — the description says the row icon, chart colour and Widget colour are `#F34E3F`, and mentions Discord only as "the Discord asset/label maps". If you want the registration tracked, the cleanest place is the same `docs/providers/<id>.md` note Unsloth used, and I am happy to add an `docs/providers/amp.md` saying the small image appears once an `amp` asset exists — same wording, same intent, without touching the guard-required key.

I have left the code as-is because removing the key would break CI on purpose. If you would rather hold the merge until the asset is up, that is entirely reasonable — just confirm and I will not touch anything further.

PR #706, opened 2026-09-16
fix(antigravity): align OAuth quota with CLI endpoint

haoyuqi:
## Summary

- Query grouped Antigravity quota summaries from the daily Cloud Code endpoint first, matching the endpoint used by the AGY CLI.
- Keep the production endpoint as a compatibility fallback for unavailable, unsupported, or empty daily responses.
- Preserve richer grouped OAuth quota when a matching local RPC source only returns an empty or legacy three-pool snapshot.
- Document the source-selection and endpoint-fallback behavior.

## Root cause

The production and daily Cloud Code quota-summary endpoints can return different Gemini windows for the same account and project. The production endpoint returned a fresh 100% five-hour window while AGY was actively consuming the daily window. Token Monitor treated the successful production response as authoritative.

A matching local RPC legacy snapshot could also replace the richer OAuth grouped quota.

### AGY CLI quota

截屏2026-09-16 下午9 58 34

### Token Monitor stable version

截屏2026-09-16 下午10 00 40

### Token Monitor with this fix

截屏2026-09-16 下午9 58 14

## Verification

- `npm run verify` in the Laradock workspace with Node.js 22.23.2
- 4,475 tests passed, 5 skipped, 0 failed
- Manual comparison confirmed that the source build matches AGY CLI values for Gemini and Claude/GPT five-hour and weekly windows


Javis603:
Thanks @haoyuqi for tracing the mismatch to the daily Cloud Code quota endpoint and for covering the local RPC interaction as well. The AGY comparison made both failure paths clear.

I verified the final head and the source-selection behavior is right: OAuth now follows the daily grouped summary with production kept as a compatibility fallback, while a matching legacy three-pool RPC snapshot no longer replaces the richer five-hour and weekly windows.

PR #756, opened 2026-09-20
fix(linux): unlock floating bubble before expanding

### LINKED ISSUE #565: [Bug][Linux] Floating Bubble gets stuck in collapsed state and cannot expand back to full UI after window blur
## Description
When running Token Monitor via AppImage on Linux with `floatingBubbleEnabled` (or default auto-collapse), switching focus away from the application collapses the window into the mini floating bubble/chart widget on the screen edge as intended. However, clicking or interacting with the collapsed widget fails to expand the full main UI back. The window remains stuck in the collapsed mini-chart/handle state, even though the background token monitoring services continue to work properly.

## Environment
- **OS**: Linux (e.g., Ubuntu 22.04 / 24.04 / Arch Linux / Fedora) 
- **Desktop Environment**: GNOME / KDE Plasma / XFCE 
- **Display Server**: Wayland / X11 
- **Token Monitor Version**: AppImage (v0.44.x / latest)

## Steps to Reproduce
1. Launch Token Monitor on Linux via AppImage.
2. Enable Floating Bubble (`floatingBubbleEnabled: true`) or keep default behavior.
3. Switch focus to another application (causing the Token Monitor window to trigger a `blur` event).
4. Observe the window auto-collapsing into the slim floating bubble / mini bar graph on the edge of the screen.
5. Click on the floating bubble tab to expand back to the main UI.

## Expected Behavior
Clicking the floating bubble should expand the window back to its normal size and render the full dashboard/UI.

## Actual Behavior
- The window remains trapped in the collapsed mini-size / chart state.
- Clicking or dragging the bubble fails to restore the main window.
- The only way to recover the main UI is to kill the process and manually edit `settings.json` to set `"floatingBubbleEnabled": false`.

## Root Cause Analysis & Technical Context
Looking into `src/electron/main.js` and `floatingBubble.js`:

1. **GTK Window Size Constraint Lock on Linux**:
   - In `collapseFloatingBubble()`, `applyCollapsedFloatingBubbleLimits()` is called, which invokes `mainWindow.setResizable(false)` and locks `setMinimumSize` / `setMaximumSize` to the bubble dimensions (`18x34`).
   - In `expandFloatingBubble()`, `restoreWindowSizeLimits()` updates the min/max limits and calls `mainWindow.setBounds(target)`. However, on Linux/GTK, calling `setBounds` while the window is marked unresizable (or before the window manager updates the size hints) causes the resize request to be ignored or clamped by window managers (Mutter / KWin).
   - On Windows, this was bypassed via `replaceMainWindow()`, but Linux currently shares the in-place resize path without unlocking `setResizable(true)` first.

2. **Focus/Blur Thrashing**:
   - On Linux (especially under Wayland / XWayland with transparent frameless windows), pointer events on the bubble do not always transition focus cleanly. When expanding, if the window does not register as focused immediately, `win.on('blur')` fires `scheduleFloatingBubbleAutoCollapse()` within 180ms, immediately collapsing it back.

## Suggested Fix
1. In `expandFloatingBubble()`, ensure `mainWindow.setResizable(true)` is explicitly applied before `mainWindow.setBounds(target)` on Linux, or use a window rebuild mechanism similar to `replaceMainWindow` if GTK constraint synchronization is unreliable.
2. Add a guard/delay on `win.on('blur')` during expansion on Linux to avoid immediate re-collapse.
3. Consider defaulting `floatingBubbleEnabled: false` on Linux until reliable window-manager resizing behavior is verified.

JJJ-NONAME:
## Summary

- unlock the in-place floating-bubble window before restoring its expanded size constraints and bounds
- keep the Windows rebuild path and the existing post-expansion window settings unchanged
- add a regression test that locks the Electron window-call order

Fixes #565

## Why

Collapsing the floating bubble marks the BrowserWindow non-resizable. The non-Windows expansion path previously restored the min/max hints and called `setBounds()` before `applyWindowSettings()` made the window resizable again. GTK window managers can ignore or clamp that resize while the non-resizable hint is still active, leaving the Linux AppImage stuck at the collapsed dimensions.

The new helper makes the required order explicit: unlock the window, restore the normal size hints, then apply the expanded bounds. The existing 300 ms collapse-suppression guard remains unchanged.

## Validation

- `npm run verify` — 4,714 passed, 0 failed, 7 skipped on Node 24.14.1
- `node --test tests/electron/floatingBubble.test.js` — 18 passed

I do not have a Linux GUI environment available for an AppImage recording, so the platform-specific behavior is covered by the BrowserWindow call-order regression test rather than a screenshot or GIF.

---
## Summary

Fixes the Linux floating-bubble expansion getting stuck at the collapsed size. Collapsing pins the window to the handle size and marks it non-resizable; Electron's Linux window then remembers those handle-sized constraints and restores them when the window becomes resizable again, which landed after `setBounds()` under the old order and clamped the bubble back to the handle.

- Adds `restoreFloatingBubbleWindow`, which unlocks the window, restores the min/max size hints, and then applies the expanded bounds in that order.
- Adds `applyWindowSizeLimits` so `restoreWindowSizeLimits()` and the new helper share one size-hint path; `restoreWindowSizeLimits()` deliberately does not unlock, since it also runs for a window that must stay non-resizable.
- Requires `limits` explicitly (no default) so an omitted pair cannot reach `setMinimumSize(undefined, undefined)`.
- Uses the helper in the non-Windows expansion path; the Windows rebuild path and post-expansion settings are unchanged.
- Adds a regression test asserting the window call order.
- Fixes #565.

| Area | Before | After |
| --- | --- | --- |
| Floating bubble expand (Linux) | Bounds applied while the window is still non-resizable; the recalled handle-size constraints land after `setBounds()` and clamp the bubble back to collapsed size | Window is made resizable before min/max hints and bounds are restored |

The 300 ms collapse-suppression guard is untouched. The helpers guard their calls (e.g., skip destroyed windows or missing `BrowserWindow` methods), so the change is safe in environments without a full `BrowserWindow`.

**Validation**

- `npm run verify` passes (4,714 passed, 0 failed, 7 skipped on Node 24.14.1).
- `node --test tests/electron/floatingBubble.test.js` passes.
- No Linux GUI was available for an AppImage recording, so Linux-specific behavior is covered by the call-order regression test.

<details>
<summary>中文版本</summary>

## 摘要

修正 Linux 浮動氣泡在展開時卡在縮小尺寸的問題。收合時，視窗會被釘在把手尺寸並標記為不可調整大小；Electron 的 Linux 視窗會記住該把手尺寸的限制，並在視窗重新變為可調整大小時還原這些限制。在舊有順序下，這會在 `setBounds()` 之後才套用，令氣泡被夾回把手尺寸。

- 新增 `restoreFloatingBubbleWindow`，依序解鎖視窗、還原最小／最大尺寸提示，然後套用展開後的邊界。
- 新增 `applyWindowSizeLimits`，讓 `restoreWindowSizeLimits()` 與新輔助函式共用同一尺寸提示路徑；`restoreWindowSizeLimits()` 刻意不解鎖視窗，因為它亦會用於必須維持不可調整大小的視窗。
- 明確要求 `limits` 參數（不設預設值），避免省略時誤傳 `setMinimumSize(undefined, undefined)`。
- 非 Windows 的展開路徑改用該輔助函式；Windows 重建路徑及展開後的設定維持不變。
- 新增回歸測試，驗證視窗呼叫的順序。
- 修正 #565。

| 項目 | 之前 | 之後 |
| --- | --- | --- |
| 浮動氣泡展開（Linux） | 視窗仍不可調整大小時套用邊界；被記住的把手尺寸限制在 `setBounds()` 之後套用，令氣泡被夾回縮小尺寸 | 先將視窗設為可調整大小，再還原最小／最大提示及邊界 |

300 毫秒的收合抑制保護維持不變。輔助函式對呼叫加入保護（例如略過已銷毀的視窗或缺少的 `BrowserWindow` 方法），因此在缺少完整 `BrowserWindow` 的環境中亦為安全。

**驗證**

- `npm run verify` 通過（Node 24.14.1，4,714 通過、0 失敗、7 略過）。
- `node --test tests/electron/floatingBubble.test.js` 通過。
- 環境中沒有 Linux GUI 可供錄製 AppImage，因此 Linux 特定行為由呼叫順序的回歸測試覆蓋。

</details>

<sup>Written for commit 6c31217f157ba2d3b0725347be8f37a957b60ae2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Javis603/token-monitor/pull/756?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Restoring a collapsed floating bubble now reliably re-enables resizing, reapplies minimum and maximum size limits, and restores its previous dimensions.
  * Window restoration now follows a consistent sequence for improved reliability.
  * Invalid or unavailable floating-bubble windows are handled safely without triggering additional window actions.


Javis603:
Thanks @JJJ-NONAME for tracking this down to the ordering and locking it with a call-order test — the sequence is untouched in what lands.

I pushed a follow-up to your branch: restored the `isDestroyed()` guard, made `limits` required instead of defaulting to `{}` (closes the CodeRabbit finding), folded `restoreWindowSizeLimits()` into the same helper, and rewrote the comment — the cause is Electron restoring the constraints it remembered at collapse time, not a window manager ignoring the resize.

